PR #5, opened 2024-06-13
refactor: separate contract interfaces from impls

fusede:


contracts/src/intent/mod.rs
```diff
@@ -35,13 +82,15 @@ impl DetailedIntent {
 
     /// Get the inner intent
     #[must_use]
-    pub const fn get_intent(&self) -> &Intent {
+    #[inline]
+    pub const fn intent(&self) -> &Intent {
```
gonchar1987, line 86: Why without `get` prefix?

contracts/src/intent/mod.rs
```diff
@@ -1,11 +1,58 @@
...
+        Intent,
+    ),
+    ExecuteIntent(
+        /// ID
+        String,
+    ),
+}
+
+impl Action {
+    pub fn decode_base64(msg: impl AsRef<[u8]>) -> Result<Self, IntentError> {
```
gonchar1987, line 48: `base64` could be confusing here from my point of view

contracts/src/account.rs
```diff
@@ -0,0 +1,23 @@
...
+    /// Return all [`Account`]s owned by given `owner`
+    fn get_accounts(&self, owner: &AccountId) -> Vec<(String, Account)>;
+
+    /// Return MPC contract for this
+    fn mpc_contract(&self) -> &AccountId;
+}
+
+#[derive(Debug, Clone, Default, Eq, PartialEq)]
+#[near(serializers=[borsh, json])]
+pub struct Account {
```
gonchar1987, line 21: I would propose promoting the module to directory and placing types to the `types.rs` and traits describe in mod.rs file:
```
-- account
    |_ mod.rs
    |_ types.rs 
```

gonchar1987 (approved):

PR #6, opened 2024-06-13
feat: add status and ttl for intent

gonchar1987:
The PR adds two additional properties to the `Intent`:

- intent status to prevent attempts to withdraw tokens more than once per intent
- minimum TTL for the intent, which doesn't allow to rollback an intent instantly after creation

### DIFF
--- a/intent/src/lib.rs
+++ b/intent/src/lib.rs
@@ -243,17 +243,15 @@ impl IntentContract {
 
         detailed_intent.set_status(Status::Processing);
 
+        let intent = detailed_intent.get_intent();
         let current_id = env::current_account_id();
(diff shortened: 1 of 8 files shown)


intent/src/lib.rs
```diff
@@ -181,6 +195,80 @@ impl IntentContract {
...
+        );
+
+        self.assert_solver(&solver_id);
+
+        let detailed_intent = self
+            .intents
+            .get_mut(id)
+            .ok_or(ContractError::IntentNotFound)?;
+
+        if !matches!(detailed_intent.get_status(), Status::Created) {
```
fusede, line 240: Can we rename it here to `Status::Available` for better readability?

fusede (approved):

PR #7, opened 2024-06-25
fix: storage deposit issue

gonchar1987:


intent/src/lib.rs
```diff
@@ -68,12 +76,22 @@ impl IntentContract for IntentContractImpl {
             "Only initiator, self or owner can roll back the intent"
         );
 
+        require!(
+            env::prepaid_gas()
+                .checked_sub(env::used_gas())
+                .unwrap_or_default()
```
fusede, line 82:
```suggestion
                .saturating_sub(env::used_gas())
```

intent/src/lib.rs
```diff
@@ -103,16 +121,47 @@ impl FungibleTokenReceiver for IntentContractImpl {
...
-        match action {
+        let promise = match action {
             Action::CreateIntent(id, intent) => {
                 log!(
                     "Creating the intent with id: {id} by: {sender_id}, amount: {}",
                     amount.0
                 );
-                self.create_intent(id, amount, intent).unwrap()
+
+                require!(id.len() <= 128, "ID is too long");
```
fusede, line 131: good catch! 👍🏻

intent/src/lib.rs
```diff
@@ -132,20 +181,86 @@ impl IntentContractImpl {
...
+        &mut self,
+        id: String,
+        amount: U128,
+        intent: Intent,
+        #[callback_result] result: Result<Option<StorageBalance>, near_sdk::PromiseError>,
+    ) -> U128 {
+        match result {
+            Ok(Some(balance)) => {
+                assert!(
+                    balance.total >= MIN_STORAGE_DEPOSIT,
```
fusede, line 229: This check doesn't really provide any guarantees that future transfer to the user would succeed, since this constant `MIN_STORAGE_DEPOSIT` can differ between different fungible token implementations. What if this contract is optimized in some way and requires less NEAR to be deposited for storage? This check would fail and user would not be able to create an intent? Or the opposite: what if it requires more NEAR to be deposited? This check would succeed but execution of this intent would fail.
There is a more generic question: are you sure that all fungible tokens implement Storage Management (NEP-145) standard? What if creators of a ft token decided to pay for storage for their users and implement some different incentive mechanism to pay for storage in this exact token, let's say? I don't know if anyone really does it, just thinking about potential consequences

intent/src/lib.rs
```diff
@@ -267,14 +378,20 @@ impl IntentContractImpl {
...
-                .then(Self::ext(current_id).change_intent_status(id, Status::Completed))
+                .then(Self::ext(current_id).change_intent_status(id, Status::Completed, 0.into()))
         };
 
-        Ok(PromiseOrValue::Promise(promise))
+        Ok(promise)
     }
 }
+
+// Could be removed after merging https://github.com/near/near-sdk-rs/pull/1208
```
fusede, line 393: it was recently merged 👍🏻

intent/src/lib.rs
```diff
@@ -68,12 +76,22 @@ impl IntentContract for IntentContractImpl {
             "Only initiator, self or owner can roll back the intent"
         );
 
+        require!(
+            env::prepaid_gas()
+                .checked_sub(env::used_gas())
+                .unwrap_or_default()
+                >= ROLLBACK_INTENT_GAS,
```
fusede, line 83: seems like we should also take into account gas for the transfer itself?

intent/src/lib.rs
```diff
@@ -103,16 +121,47 @@ impl FungibleTokenReceiver for IntentContractImpl {
...
+
+                require!(id.len() <= 128, "ID is too long");
+
+                // First check that initiator has storage deposit on token he wants to get.
+                ext_storage_management::ext(intent.receive.token_id.clone())
+                    .storage_balance_of(sender_id)
+                    .then(
+                        Self::ext(env::current_account_id())
+                            .with_static_gas(FINISH_CREATING_GAS)
+                            .finish_creating_intent(id, amount, intent),
```
fusede, line 139: maybe check in advance that we don't have an intent with such id? or create it with locked status?

intent/src/lib.rs
gonchar1987: This intent covers NEP-141 tokens, which have to implement NEP-145 as well. What you mentioned is out of the scope of this intent.

intent/src/lib.rs
gonchar1987: But we do [it](https://github.com/defuse-is-near/defuse-contracts/blob/fix/storage_deposit/intent/src/lib.rs#L332) in the `create_intent` method.

fusede: `create_intent` is called only in `finish_creating_intent`. By making this check only after response from `storage_balance_of` we just spend user's gas while we can fail early

mitinarseny (approved):

fusede (approved):

PR #9, opened 2024-07-01
fix: concurrent solvers

gonchar1987:


contracts/src/intent/error.rs
```diff
@@ -5,24 +5,18 @@ use thiserror::Error as ThisError;
 pub enum IntentError {
     #[error("Borsh")]
     Borsh,
-    #[error("base64: {0}")]
+    #[error("Base64: {0}")]
```
fusede, line 8: Hm, is there a convention for error messages in NEAR?
According to Rust guidelines, `Display` implementations for error messages [should](https://github.com/rust-lang/api-guidelines/blob/master/src/interoperability.md#error-types-are-meaningful-and-well-behaved-c-good-err) start with lowercase letter, so they can be then combined chained together in a more nice format

intent/src/lib.rs
```diff
@@ -236,12 +236,14 @@ impl IntentContractImpl {
...
-        detailed_intent: &mut DetailedIntent,
         solver_id: &AccountId,
         #[callback_result] result: Result<Option<StorageBalance>, near_sdk::PromiseError>,
     ) -> PromiseOrValue<U128> {
         match result {
-            Ok(Some(_)) => self.execute_intent(id, amount, detailed_intent).unwrap(),
+            Ok(Some(_)) => match self.execute_intent(id, amount) {
+                Ok(promise) => promise,
+                Err(e) => env::panic_str(&e.to_string()),
+            },
```
fusede, line 246:
```suggestion
            Ok(Some(_)) => self.execute_intent(id, amount).unwrap(),
```

fusede (changes requested):

mitinarseny (approved):

fusede (approved):

PR #11, opened 2024-07-23
feat: cross chain swap

fusede:


tests/src/tests/intents/swap_ft/env.rs
```diff
@@ -45,6 +45,18 @@ impl Env {
             .set_min_ttl(self.intent.id(), min_ttl)
             .await;
     }
+
+    pub async fn ft_mint(
+        &self,
+        token: &AccountId,
```
gonchar1987, line 51: `ft_mint` calls could be written in one line again in the tests if you change the type of the argument to `&Contract` 😄

intents/swap/src/lib.rs
```diff
@@ -210,67 +213,56 @@ impl SwapIntentContractImpl {
...
-            // ft_on_transfer()
-            Asset::Ft(FtAmount { amount, .. }) => {
-                PromiseOrValue::Value(json!(if refund { *amount } else { 0.into() }))
-            }
-            // nft_on_transfer()
-            Asset::Nft(_) => PromiseOrValue::Value(json!(refund)),
+                // nft_on_transfer()
+                NearAsset::Nep171(_) => PromiseOrValue::Value(json!(refund)),
+            },
+            // on_cross_chain_tranfer()
```
gonchar1987, line 260:
```suggestion
            // on_cross_chain_transfer()
```

intents/swap/src/lib.rs
```diff
@@ -22,7 +24,7 @@ mod rollback;
 #[near(contract_state)]
 #[derive(PanicOnDefault)]
 pub struct SwapIntentContractImpl {
-    intents: LookupMap<IntentId, Mutex<SwapIntentStatus>>,
+    intents: LookupMap<IntentId, Mutex<SwapIntent>>,
```
gonchar1987, line 27: Does it make sense to change the `LookupMap` with the iterable `UnorderedMap` to allow to do garbage collection in the future? Also, such change will require making data migration, which is always fun 😄

contracts/src/intents/swap/asset.rs
```diff
@@ -1,53 +1,157 @@
...
+
+    #[must_use]
+    #[inline]
+    pub fn with_account(&self, account: GenericAccount) -> Option<Self> {
+        Self::new(self.asset(), account)
+    }
+
+    #[must_use]
+    #[inline]
+    pub const fn gas_for_transfer(&self) -> Gas {
```
gonchar1987, line 75: Looks like the method isn't used.

intents/swap/src/cross_chain.rs
```diff
@@ -0,0 +1,61 @@
...
+                amount,
+            },
+        };
+        Ok(match action {
+            SwapIntentAction::Create(create) => {
+                self.create_intent(received, create)?;
+                PromiseOrValue::Value(true)
+            }
+            SwapIntentAction::Execute(execute) => self.execute_intent(&received, execute)?.into(),
+        })
```
gonchar1987, line 48: Mayby more idiomatic would be:
```rust
match action { 
    SwapIntentAction::Create(create) => self
         .create_intent(received, create)
         .map(|_| PromiseOrValue::Value(true)),
    SwapIntentAction::Execute(execute) => {
         self.execute_intent(&received, execute).map(Into::into)
    }
}
```

intents/swap/src/cross_chain.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// `msg` parameter should contain [`SwapIntentAction`] serialized to
+    /// JSON string.
+    fn cross_chain_on_transfer(
+        &mut self,
+        asset: String,
+        amount: String,
+        msg: String,
+    ) -> PromiseOrValue<bool> {
+        self.internal_cross_chain_action(asset, amount, msg)
+            .unwrap()
```
gonchar1987, line 20: `env::panic_str`?

intents/swap/src/lost_found.rs
```diff
@@ -17,15 +14,18 @@ impl LostFound for SwapIntentContractImpl {
```
gonchar1987, line 14: In the method `lost_found` replace:
`assert_eq!(env::attached_deposit(), NearToken::from_yoctonear(1));`
to 
`near_sdk::assert_one_yocto();`
And as I know they don't recommend using `assert`s in the contract code. There is analog: `near_sdk::require!`.

gonchar1987 (commented):
Overall looks good. Good job 👍

tests/src/tests/intents/swap_ft/env.rs
fusede: You're right, but in some places it's still required to be able to use with `#[rstest]`:
https://github.com/defuse-protocol/defuse-contracts/blob/72bf146a9479d85b793ebc8357343aaf3fbc3e71/tests/src/tests/intents/swap/mod.rs#L78-L88

intents/swap/src/lib.rs
fusede: Good point, I'll fix it in a separate PR as we discussed

gonchar1987 (approved):
👍

PR #17, opened 2024-10-15
Add wasm toolkit that serializes some structs using borsh

cawabunga-bytes:


cawabunga-bytes:
Closing this, because we decided to get rid of borsh serialization and wasm is not needed anymore.

PR #49, opened 2025-02-26
feat(intents): new storage_deposit intent

fusede:


gonchar1987 (approved):
Do it make sense to add some integration tests for this?

core/src/intents/mod.rs
```diff
@@ -44,6 +44,8 @@ pub enum Intent {
     MtWithdraw(MtWithdraw),
     NativeWithdraw(NativeWithdraw),
 
+    StorageDeposit(StorageDeposit),
```
BonelessImpl, line 47: I believe this can be a problem in Borsh. If Borsh uses the index of the arm for encoding, i.e., the serialized data is the index followed by the data inside the arm, this will break backwards compatibility where deserializing previous versions will fail.

If borsh is never used (and will be used), then we should remove it and drop its support until there is intent to properly support it. Having Borsh and ignoring its backwards compatibility doesn't sound like the right thing to do.

core/src/engine/state/mod.rs
```diff
@@ -137,4 +137,18 @@ pub trait State: StateView {
...
+        owner_id: &AccountIdRef,
+        storage_deposit: StorageDeposit,
+    ) -> Result<()> {
+        self.internal_withdraw(
+            owner_id,
+            [(
+                TokenId::Nep141(self.wnear_id().into_owned()),
+                storage_deposit.amount.as_yoctonear(),
+            )],
+        )
```
BonelessImpl, line 152: I wonder whether it's better to just make `execute_intents()` payable... I see the appeal here, where we plan everything in advanced... I'm not sure. The downside I see is that the user must have the wnear already in his account for this to work. At best, we enforce adding a conversion step from near to wnear.

defuse/src/contract/tokens/nep141/storage_deposit.rs
```diff
@@ -0,0 +1,23 @@
+use defuse_core::intents::tokens::StorageDeposit;
+use near_contract_standards::storage_management::ext_storage_management;
+use near_sdk::{env, near, require, Gas, Promise, PromiseResult};
+
+use crate::contract::{tokens::STORAGE_DEPOSIT_GAS, Contract, ContractExt};
+
+#[near]
+impl Contract {
```
BonelessImpl, line 8: I have to say that sprinkling `impl Contract` over many modules is making it very hard to track how Contract works... maybe we should use some delegation to call these functions, and maintain `impl Contract` strictly being in one file.

core/src/intents/tokens.rs
```diff
@@ -197,3 +197,31 @@ impl ExecutableIntent for NativeWithdraw {
...
 }
+
+/// Make [NEP-145](https://nomicon.io/Standards/StorageManagement#nep-145)
+/// `storage_deposit` for an `account_id` on `contract_id`.
+/// The `amount` will be subtracted from user's NEP-141 `wNEAR` balance.
+/// NOTE: the `wNEAR` will not be refunded in any case.
+#[near(serializers = [borsh, json])]
+#[derive(Debug, Clone)]
+pub struct StorageDeposit {
+    pub contract_id: AccountId,
```
BonelessImpl, line 208: Do we really need this? Wouldn't be feasible to use `env::predecessor_account_id()` or something similar in the external call?

defuse/src/contract/intents/state.rs
```diff
@@ -177,4 +177,33 @@ impl State for Contract {
...
+        owner_id: &AccountIdRef,
+        storage_deposit: StorageDeposit,
+    ) -> Result<()> {
+        self.withdraw(
+            owner_id,
+            [(
+                TokenId::Nep141(self.wnear_id().into_owned()),
+                storage_deposit.amount.as_yoctonear(),
+            )],
+            Some("withdraw"),
```
BonelessImpl, line 192: After the last PR(s), we should probably call this "subtract"

core/src/engine/state/mod.rs
```diff
@@ -137,4 +137,18 @@ pub trait State: StateView {
             )],
         )
     }
+
+    fn storage_deposit(
+        &mut self,
+        owner_id: &AccountIdRef,
+        storage_deposit: StorageDeposit,
+    ) -> Result<()> {
```
BonelessImpl, line 145: Are default impls for this trait necessary?

BonelessImpl (approved):

BonelessImpl (commented):
Perhaps this should be rebased on master before merging. I'm sure there will be merge conflicts that need manual fixing.

core/src/intents/mod.rs
mitinarseny: Agreed, removing borsh for now would be a good choice to prevent devs from using it

core/src/intents/tokens.rs
fusede: This intent is intended (sorry) to be used to make storage_deposit on any other smart-contract specified by user.
How else would you know which contract_id this user is willing to make storage_deposit to?

core/src/engine/state/mod.rs
fusede: Not really, I've used to just as indication what should be done in the real implementation. But I agree, that it can lead to mistakes, so better remove it

core/src/intents/mod.rs
BonelessImpl: Done

tests/src/tests/defuse/storage/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+#[rstest]
+#[trace]
+#[case(
+    Seed::from_entropy(),
+    MIN_FT_STORAGE_DEPOSIT_VALUE,
+    Some(MIN_FT_STORAGE_DEPOSIT_VALUE)
+)]
+#[case(
+    Seed::from_entropy(),
+    NearToken::from_yoctonear(1_250_000_000_000_000_000_000-1), // Sending less than the required min leads to nothing being deposited
```
fusede, line 25:
```suggestion
    MIN_FT_STORAGE_DEPOSIT_VALUE.saturating_sub(NearToken::from_yoctonear(1)), // Sending less than the required min leads to nothing being deposited
```

tests/src/tests/defuse/storage/mod.rs
```diff
@@ -0,0 +1,234 @@
...
+    Seed::from_entropy(),
+    NearToken::from_yoctonear(1_250_000_000_000_000_000_000+1),
+    Some(MIN_FT_STORAGE_DEPOSIT_VALUE)
+)]
+async fn storage_deposit_success(
+    #[case] seed: Seed,
+    #[case] amount_to_deposit: NearToken,
+    #[case] expected_deposited: Option<NearToken>,
+) {
+    let mut rng = make_seedable_rng(seed);
```
fusede, line 38: Does it make sense to use [`fixture`](https://docs.rs/rstest/latest/rstest/attr.fixture.html#) here? So `make_seedable_rng` will be a fixture, while `rng` will become a parameter of this test function

BonelessImpl: I added using fixtures. Seems simple enough. Thanks for the suggestion.

BonelessImpl: Btw, we can't put make_seedable_rng in the fixture, because we need the inner seed that creates all the randomness of the test.

core/src/tokens.rs
```diff
@@ -52,6 +52,17 @@ pub enum TokenId {
     ),
 }
 
+impl TokenId {
+    pub fn account_id(&self) -> &AccountIdRef {
```
fusede, line 56: It seems like it should be named `contract_id()`, but we might introduce new `TokenId` variants in the future, which will be minted/burned entirely within our contract and does not interact with any external contract, so I'd remove it completely

tests/src/tests/defuse/env.rs
```diff
@@ -101,6 +101,32 @@ impl Env {
...
+    pub async fn near_balance(&mut self, account_id: &AccountId) -> NearToken {
+        self.sandbox
+            .worker()
+            .view_account(account_id)
+            .await
+            .unwrap()
+            .balance
+    }
+
+    pub fn sandbox_ref(&self) -> &Sandbox {
```
fusede, line 123:
```suggestion
    pub fn sandbox(&self) -> &Sandbox {
```

tests/src/tests/defuse/intents/ft_withdraw.rs
```diff
@@ -24,12 +24,11 @@ use crate::{
...
 #[rstest]
 #[trace]
-#[case(Seed::from_entropy())]
-async fn test_ft_withdraw_intent(#[case] seed: Seed) {
+async fn test_ft_withdraw_intent(random_seed: Seed) {
     // intentionally large deposit
     const STORAGE_DEPOSIT: NearToken = NearToken::from_near(1000);
 
-    let mut rng = make_seedable_rng(seed);
+    let mut rng = make_seedable_rng(random_seed);
```
fusede, line 31: I mean, maybe use `make_seedable_rng` itself as a fixture?

tests/src/tests/defuse/storage/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+        let storage_balance_ft1_user2 = env
+            .storage_balance_of(&env.ft1, env.user2.id())
+            .await
+            .unwrap();
+
+        assert!(storage_balance_ft1_user1.is_none());
+        assert!(storage_balance_ft1_user2.is_none());
+    }
+
+    // For intents contract to have a balance in wnear, we make a storage deposit for it
```
fusede, line 63: Move it to `Env::build()`?

tests/src/tests/defuse/storage/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    // The user should have enough wnear in his account (in his account in the wnear contract)
+    env.user2
+        .near_deposit(env.wnear.id(), NearToken::from_near(100))
+        .await
+        .unwrap();
+
+    // Fund the user's account with near in the intents contract for the storage deposit intent
+    env.defuse_ft_mint(
+        env.wnear.id(),
```
fusede, line 102: How are you "minting" wNEAR? This should probably use a different method, not `defuse_ft_mint`

tests/src/tests/defuse/storage/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    let intents = DefuseIntents {
+        intents: [storage_deposit_intent].into(),
+    };
+
+    let signed_intents = env.user2.sign_defuse_message(
+        env.defuse.id(),
+        rng.random(),
+        Deadline::timeout(std::time::Duration::from_secs(120)),
+        intents,
```
fusede, line 124: It's hard to read when these values are spread over the code. Inlining them is more readable:
```suggestion
        DefuseIntents {
           intents: [StorageDeposit {
               contract_id: env.ft1.clone(),
               account_id: env.user2.id().clone(),
               amount: amount_to_deposit,
           }.into()]
         .into(),
       }
```

tests/src/tests/defuse/storage/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+    );
+
+    env.defuse.execute_intents([signed_intents]).await.unwrap();
```
fusede, line 127: Even better to inline them here, as you don't use these values in the code below anyway

tests/src/tests/defuse/storage/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+    };
+
+    let signed_intents = env.user2.sign_defuse_message(
+        env.defuse.id(),
+        rng.random(),
+        Deadline::timeout(std::time::Duration::from_secs(120)),
+        intents,
+    );
+
+    // Fails the user does not own any near in the intents smart contract. They should first deposit near/wnear.
```
fusede, line 227:
```suggestion
    // Fails the user does not own any wNEAR in the intents smart contract. They should first deposit wNEAR.
```

tests/src/utils/ft.rs
```diff
@@ -38,37 +42,33 @@ pub trait FtExt: StorageManagementExt {
...
+
     async fn ft_storage_deposit(
         &self,
         token_id: &AccountId,
         account_id: Option<&AccountId>,
-    ) -> anyhow::Result<StorageBalance> {
-        self.storage_deposit(token_id, account_id, FT_STORAGE_DEPOSIT)
-            .await
-    }
+    ) -> anyhow::Result<StorageBalance>;
```
fusede, line 50: Why to remove it? This method does not depend on the implementation, the same way how `core::iter::Iterator` only requires `next()` to be defined, and other methods have default implementations

tests/src/utils/storage_management.rs
```diff
@@ -54,6 +61,23 @@ impl StorageManagementExt for near_workspaces::Account {
             .json()
             .map_err(Into::into)
     }
+
+    async fn storage_balance_of(
+        &self,
+        contract_id: &AccountId,
+        account_id: &AccountIdRef,
+    ) -> anyhow::Result<Option<StorageBalance>> {
+        self.call(contract_id, "storage_balance_of")
```
fusede, line 70: `storage_balance_of` is a view method, you don't need to send a transaction, just use `.view()`

tests/src/tests/defuse/storage/mod.rs
BonelessImpl: Actually I have the exact opposite view... cramming code makes it harder for me to read. In fact, this was difficult for me when I first started reading the code. 

If you insist, I'd do it, but I'm not particularly happy about it.

One thing I can suggest is wrapping these in one scope, like I did above. This way, it prevents mixing them with other, unrelated things.

tests/src/tests/defuse/storage/mod.rs
BonelessImpl: I'll resolve this to keep the discussion in one comment.

tests/src/tests/defuse/storage/mod.rs
fusede: My reasoning for inlining here is that it should be as close as possible to JSON representation of the signed intent, that will be submitted on-chain, e.g.:
```json
{
  "signed": [
    {
      "payload": "{\"signer_id\":\"465c5516b703bc8a925f8e0f4a47337f69302be899050f2904bd5dce5f07afde\",\"verifying_contract\":\"intents.near\",\"deadline\":\"2025-03-20T09:54:34.684Z\",\"nonce\":\"4dusNOOVIVXuB7z6F8LAOCKXfEGvMBiYm5nlH4ltTg0=\",\"intents\":[{\"intent\":\"ft_withdraw\",\"token\":\"sol-5ce3bf3a31af18be40ba30f721101b4341690186.omft.near\",\"receiver_id\":\"sol-5ce3bf3a31af18be40ba30f721101b4341690186.omft.near\",\"amount\":\"4000000\",\"memo\":\"WITHDRAW_TO:5jfCVJ7oZ2Y38Nu28qCdRuY5KgsuqRKZtwYzCY7q3DNV\"}]}",
      "public_key": "ed25519:5jfCVJ7oZ2Y38Nu28qCdRuY5KgsuqRKZtwYzCY7q3DNV",
      "signature": "ed25519:3udGxDpiVwSjA7bNJbCfqZ7S72HffvSvRY9vG9tEcbZbinNotRfipw65AXUpCtq4FyxHvpyKuk1cFynQsXTnPKx5",
      "standard": "raw_ed25519"
    }
  ]
}
```

BonelessImpl: I don't love it... but alright. Done.

tests/src/tests/defuse/storage/mod.rs
BonelessImpl: I'm not sure I understand... the fact that everything is prepared in env without qualifiers makes the tests extremely hard to read and error prone, where we add things that not every test needs... tests should be as isolated as possible, and anything common should be built by pre-prepared and well-defined fixtures. The Env is a great idea, but it should be as minimal as possible and have no state in it after construction. I would love to fix this and just remove all the default preparation and put them behind guards that need to be manually activated in the builder.

tests/src/tests/defuse/storage/mod.rs
BonelessImpl: Well, I looked at the impl, and it's just deposit... what do you suggest?

fusede: I see. Maybe rename to `defuse_ft_deposit()`?

tests/src/tests/defuse/intents/ft_withdraw.rs
BonelessImpl: This can be done with some trait magic to make it possible to view the seed value on error automatically, but the problem is that recreating the rng to replay the test with a given seed will require reimporting all used Rng trait primitives, since the type that implements the Rng cannot be contained implicitly in the trait that represents the generator object (which is a super trait of Rng and CryptoRng), making the process less smooth... How enthusiastic are you about this? I wouldn't want to change this.

My changes are stashed, if you'd like to take a look.

fusede: Ok, agreed, let's keep it simple

tests/src/tests/defuse/storage/mod.rs
BonelessImpl: There's another private function with the same name... I renamed the function to defuse_ft_deposit_to

tests/src/tests/defuse/intents/ft_withdraw.rs
BonelessImpl: Cool. Btw, when you're happy and don't need me to have further attention to the comment, feel free to resolve the comment.

tests/src/utils/ft.rs
BonelessImpl: OK. I'll put it back.

PR #51, opened 2025-03-06
Fix borsh deserialization of pips and a few minor other fixes

BonelessImpl:
Currently, the invariant of `Pips` (with a max value) can be broken through bad binary data. This fixes that.

### DIFF
--- a/core/src/fees.rs
+++ b/core/src/fees.rs
@@ -215,10 +215,11 @@ pub struct FeeCollectorChangedEvent<'a> {
 impl BorshDeserialize for Pips {
     fn deserialize_reader<R: std::io::Read>(reader: &mut R) -> std::io::Result<Self> {
         let pips: u32 = near_sdk::borsh::BorshDeserialize::deserialize_reader(reader)?;
-        let pips = Self::from_pips(pips).ok_or(std::io::Error::new(
-            std::io::ErrorKind::InvalidData,
-            format!("Invalid pips value: {pips}"),
-        ))?;
-        Ok(pips)
+        Self::from_pips(pips).ok_or_else(|| {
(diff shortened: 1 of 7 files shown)


core/src/fees.rs
```diff
@@ -205,3 +211,14 @@ pub struct FeeCollectorChangedEvent<'a> {
     pub old_fee_collector: Cow<'a, AccountIdRef>,
     pub new_fee_collector: Cow<'a, AccountIdRef>,
 }
+
+impl BorshDeserialize for Pips {
+    fn deserialize_reader<R: std::io::Read>(reader: &mut R) -> std::io::Result<Self> {
+        let pips: u32 = near_sdk::borsh::BorshDeserialize::deserialize_reader(reader)?;
+        let pips = Self::from_pips(pips).ok_or(std::io::Error::new(
+            std::io::ErrorKind::InvalidData,
+            format!("Invalid pips value: {pips}"),
```
fusede, line 220: Please, try to reuse `PipsOutOfRange` error here

core/src/fees.rs
```diff
@@ -205,3 +211,14 @@ pub struct FeeCollectorChangedEvent<'a> {
...
+        ))?;
```
fusede, line 221: Use it as a return value

fusede (changes requested):

core/src/fees.rs
TheQuantumPhysicist: Cannot be done. But used to_string(). That's the closest.

BonelessImpl: Test

fusede (approved):

PR #54, opened 2025-03-07
Some renames of deposit -> add and withdraw -> subtract

BonelessImpl:


core/src/tokens.rs
```diff
@@ -128,19 +128,19 @@ impl<T> TokenAmounts<T> {
...
 
-impl<T> TokenAmounts<T>
+impl<T> Amounts<T>
 where
     T: DefaultMap,
     T::V: Copy,
 {
     #[inline]
-    pub fn balance_of(&self, k: &T::K) -> T::V {
+    pub fn amount_for(&self, k: &T::K) -> T::V {
```
fusede, line 137:
```suggestion
    pub fn get(&self, k: &T::K) -> T::V {
```

fusede (changes requested):

fusede (approved):

PR #59, opened 2025-03-20
Support omni-bridge token wrapping

BonelessImpl:
We support omni bridge by wrapping its token with the PoA tokens. This PR supports migrating a token from being used as an owned token, vs being used to wrap omni bridge.

The upgrade takes two steps:

1. Mandatory initial contract upgrade, which will add a field that contains the target wrapped token as options

2. After step 1, the token remains as PoA with nothing changes. Once the migration to Omni bridge is done, we set the wrapped token address, which will change the behavior of the smart contract to wrap omni-bridge


poa-token/src/contract.rs
```diff
@@ -108,7 +253,30 @@ impl FungibleTokenCore for Contract {
         memo: Option<String>,
         msg: String,
     ) -> PromiseOrValue<U128> {
-        self.token.ft_transfer_call(receiver_id, amount, memo, msg)
+        let Some(wrapped_token_id) = self.wrapped_token().map(|id| id.to_owned()) else {
```
fusede, line 256:
```suggestion
        let Some(wrapped_token_id) = self.wrapped_token().cloned() else {
```

poa-token/src/contract.rs
```diff
@@ -56,6 +78,115 @@ impl Contract {
...
+                }
+            }
+        };
+
+        self.token.internal_deposit(account_id, unused_amount);
+    }
+}
+
+impl Contract {
+    fn do_unwrap_and_transfer(
```
fusede, line 167: There is no preprocessing step here, so `do_` prefix is not needed
```suggestion
    fn unwrap_and_transfer(
```

poa-token/src/contract.rs
```diff
@@ -56,6 +78,115 @@ impl Contract {
...
+
+        Self {
+            token: old_state.token,
+            metadata: old_state.metadata,
+            wrapped_token: None,
+        }
+    }
+
+    #[only(self, owner)]
+    pub fn clear_wrapped_token_account_id(&mut self) {
```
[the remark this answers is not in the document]
BonelessImpl: As discussed yesterday, this is a decision to be made. Seems like we have an answer. Done.

poa-token/src/contract.rs
```diff
@@ -56,6 +78,115 @@ impl Contract {
...
+        self.wrapped_token = None;
+    }
+
+    #[only(self, owner)]
+    pub fn set_wrapped_token_account_id(&mut self, token_account_id: AccountId) -> Promise {
```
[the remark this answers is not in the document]
BonelessImpl: Done.

poa-token/src/contract.rs
```diff
@@ -56,6 +78,115 @@ impl Contract {
         .emit();
         contract
     }
+
+    #[private]
```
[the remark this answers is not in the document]
BonelessImpl: It's not possible to do this here. It doesn't compile.

poa-token/src/contract.rs
```diff
@@ -56,6 +78,115 @@ impl Contract {
...
+    }
+
+    #[only(self, owner)]
+    pub fn set_wrapped_token_account_id(&mut self, token_account_id: AccountId) -> Promise {
+        if self.wrapped_token.is_some() {
+            env::panic_str("Wrapped token is already set")
+        }
+
+        ext_ft_core::ext(token_account_id.clone())
+            .ft_balance_of(CURRENT_ACCOUNT_ID.clone())
```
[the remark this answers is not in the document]
BonelessImpl: I had already tried this

error[E0599]: no method named `with_static_gas` found for struct `near_sdk::Promise` in the current scope

poa-token/src/contract.rs
```diff
@@ -56,6 +78,115 @@ impl Contract {
...
+                    env::panic_str(&format!(
+                        "Setting token id {token_account_id} for contract {} failed due to bad balance bytes",
+                        &*CURRENT_ACCOUNT_ID
+                    ))
+                }
+            }
+            near_sdk::PromiseResult::Failed => env::panic_str(
+                "Setting token id {token_account_id} for contract {} failed due to failed promise",
+            ),
+        };
```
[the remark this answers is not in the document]
BonelessImpl: It doesn't return Result, but PromiseResult, which doesn't support functional methods like ok().

poa-token/src/contract.rs
```diff
@@ -56,6 +78,115 @@ impl Contract {
...
+}
+
+impl Contract {
+    fn do_unwrap_and_transfer(
+        &mut self,
+        token_destination: AccountId,
+        amount: U128,
+        memo: Option<String>,
+        msg: String,
+        wrapped_token_id: AccountId,
```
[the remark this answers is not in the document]
BonelessImpl: I was planning to make it an associated function. Will remove.

poa-token/src/contract.rs
```diff
@@ -82,6 +217,13 @@ impl PoaFungibleToken for Contract {
     }
 }
 
+#[near]
+impl CanWrapToken for Contract {
+    fn wrapped_token(&self) -> Option<&AccountIdRef> {
```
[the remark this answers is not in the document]
BonelessImpl: Done, even though I don't like this...

poa-token/src/contract.rs
```diff
@@ -71,6 +202,10 @@ impl PoaFungibleToken for Contract {
     #[only(self, owner)]
     #[payable]
     fn ft_deposit(&mut self, owner_id: AccountId, amount: U128, memo: Option<String>) {
+        if self.wrapped_token.is_some() {
+            env::panic_str("This PoA token was migrated to OmniBridge. No deposits are possible.");
+        }
```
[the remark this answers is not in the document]
BonelessImpl: The sync problem is probably what I'll be doing next. 

I get that you don't want ft_deposit... but since we want to support both states (with and without wrapping), I don't think there's a way to remove this function. We can do it in the future if we mandate the presence of a token, and then we make the wrapped token a non-optional value, and then we can remove this.

So far we decided to have `Option<wrapper>`... it's far simpler. I think the panic is good enough for now.

IF we decide to change this whole code and not make the wrapping optional, this will mean we have to branch our code, where the code with and without wrapping have to be maintained separately (because maybe non-migrated and deployed PoA token has a problem that needs to be fixed). If you insist on removing the optionality, then that's a discussion to be had.

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+                Self::ext(CURRENT_ACCOUNT_ID.clone())
+                    .with_static_gas(METADATA_SET_TOKEN_GAS)
+                    .with_attached_deposit(env::attached_deposit())
+                    .do_sync_wrapped_metadata(caller_id),
+            )
+    }
+
+    #[private]
+    #[payable]
+    pub fn do_sync_wrapped_metadata(&mut self, caller_id: AccountId) -> Promise {
```
mitinarseny, line 200:
```suggestion
    pub fn do_sync_wrapped_metadata(&mut self, caller_id: AccountId) -> PromiseOrValue<()> {
```

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+                    env::attached_deposit().as_yoctonear()
+                ))
+            });
+
+        near_sdk::log!(
+            "Updated metadata with cost from size {}yNEAR to size {}yNEAR with cost {}yNEAR",
+            initial_storage_usage,
+            end_storage_usage,
+            new_storage_cost
+        );
```
mitinarseny, line 257: I've suggested fixing the naming above exactly for this reason 😃

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+        let end_storage_usage = env::storage_usage();
+
+        // Note that we use saturating sub here to prevent abuse. We do not refund Near for freed storage
+        let new_storage_byte_count = end_storage_usage.saturating_sub(initial_storage_usage);
+
+        let new_storage_cost = env::storage_byte_cost()
+            .checked_mul(u128::from(new_storage_byte_count))
+            .expect("Storage cost calculation overflow");
+
+        let storage_cost_refund = env::attached_deposit()
```
mitinarseny, line 242:
```suggestion
        let refund = env::attached_deposit()
```

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+            "Wrapped token is already set"
+        );
+
+        ext_ft_core::ext(token_account_id.clone())
+            .with_static_gas(BALANCE_OF_GAS)
+            .ft_balance_of(CURRENT_ACCOUNT_ID.clone())
+            .then(
+                Self::ext(CURRENT_ACCOUNT_ID.clone())
+                    .with_static_gas(DO_WRAP_TOKEN_GAS)
+                    .do_set_wrapped_token_account_id(token_account_id),
```
mitinarseny, line 278: What if the balance changes during this async promises? We need to "pause" the contract first..

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+            new_storage_cost
+        );
+
+        Promise::new(caller_id).transfer(storage_cost_refund)
+    }
+
+    #[only(self, owner)]
+    #[payable]
+    pub fn set_wrapped_token_account_id(&mut self, token_account_id: AccountId) -> Promise {
+        assert_one_yocto();
```
mitinarseny, line 265: This method will write `Some(token_account_id)` to the storage eventually, which will result in locking more native NEAR for storage staking. We should require the caller to pay for this storage

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+
+        *self.wrapped_token_mut() = Some(token_account_id);
+    }
+
+    /// Returns the amount of tokens that were used/unwrapped after requesting an unwrap
+    #[private]
+    pub fn ft_resolve_unwrap(
+        &mut self,
+        sender_id: &AccountId,
+        amount: u128,
```
mitinarseny, line 322:
```suggestion
        amount: U128,
```

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+                    // read result of `ft_on_transfer` due to insufficient gas
+                    amount
+                } else {
+                    0
+                }
+            }
+        };
+
+        let to_refund = amount.saturating_sub(used);
+        self.token_mut().internal_deposit(sender_id, to_refund);
```
mitinarseny, line 346: Internal deposit doesn't emit `ft_mint` event by itself

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+        let Some(wrapped_token_id) = self.wrapped_token().cloned() else {
+            env::panic_str("Unwrapping is only for wrapped tokens");
+        };
+
+        let previous_owner = &*PREDECESSOR_ACCOUNT_ID;
+
+        self.ft_withdraw(previous_owner, amount, None);
+
+        // There seems to be a bug in clippy... this isn't a single match arm
+        #[allow(clippy::single_match_else)]
```
mitinarseny, line 369: clippy tells `if let` would [suffice](https://rust-lang.github.io/rust-clippy/master/index.html#single_match_else)

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+        self.token_mut().internal_deposit(sender_id, to_refund);
+
+        used.into()
+    }
+}
+
+impl Contract {
+    fn unwrap_and_transfer(
+        &mut self,
+        token_destination: AccountId,
```
mitinarseny, line 355:
```suggestion
        receiver_id: AccountId,
```

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+
+        // There seems to be a bug in clippy... this isn't a single match arm
+        #[allow(clippy::single_match_else)]
+        match msg {
+            Some(inner_msg) => ext_ft_core::ext(wrapped_token_id.clone())
+                .with_attached_deposit(NearToken::from_yoctonear(1))
+                .ft_transfer_call(token_destination, amount, memo, inner_msg)
+                .then(
+                    Contract::ext(CURRENT_ACCOUNT_ID.clone())
+                        .with_static_gas(FT_UNWRAP_GAS)
```
mitinarseny, line 376:
```suggestion
                        .with_static_gas(FT_RESOLVE_UNWRAP_GAS)
```

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+        token_destination: AccountId,
+        amount: U128,
+        memo: Option<String>,
+        msg: Option<String>,
+    ) -> PromiseOrValue<U128> {
+        let Some(wrapped_token_id) = self.wrapped_token().cloned() else {
+            env::panic_str("Unwrapping is only for wrapped tokens");
+        };
+
+        let previous_owner = &*PREDECESSOR_ACCOUNT_ID;
```
mitinarseny, line 364:
```suggestion
        let sender_id = &*PREDECESSOR_ACCOUNT_ID;
```

poa-token/src/contract.rs
```diff
@@ -56,6 +153,236 @@ impl Contract {
...
+                .ft_transfer_call(token_destination, amount, memo, inner_msg)
+                .then(
+                    Contract::ext(CURRENT_ACCOUNT_ID.clone())
+                        .with_static_gas(FT_UNWRAP_GAS)
+                        .ft_resolve_unwrap(previous_owner, amount.0, true),
+                )
+                .into(),
+            None => ext_ft_core::ext(wrapped_token_id.clone())
+                .with_attached_deposit(NearToken::from_yoctonear(1))
+                .ft_transfer(token_destination, amount, memo)
```
mitinarseny, line 382: Seems like you forget the callback `ft_resolve_unwrap(sender_id, amount, false)`

poa-token/src/contract.rs
```diff
@@ -65,14 +392,21 @@ impl PoaFungibleToken for Contract {
...
     }
 
     #[only(self, owner)]
     #[payable]
     fn ft_deposit(&mut self, owner_id: AccountId, amount: U128, memo: Option<String>) {
-        self.token.storage_deposit(Some(owner_id.clone()), None);
-        self.token.internal_deposit(&owner_id, amount.into());
+        if self.wrapped_token().is_some() {
+            env::panic_str("This PoA token was migrated to OmniBridge. No deposits are possible.");
+        }
```
mitinarseny, line 403:
```suggestion
        require!(
            self.wrapped_token().is_none(),
            "This PoA token was migrated to OmniBridge. No deposits are possible.",
        );
```

poa-token/src/contract.rs
```diff
@@ -82,6 +416,13 @@ impl PoaFungibleToken for Contract {
     }
 }
 
+#[near]
+impl CanWrapToken for Contract {
+    fn wrapped_token(&self) -> Option<&AccountId> {
```
mitinarseny, line 421: It doesn't make sense to duplicate this method across two impls

poa-token/src/contract.rs
```diff
@@ -108,15 +454,53 @@ impl FungibleTokenCore for Contract {
...
+                "No wrapping token in contract. Using legacy ft_transfer_call path. Caller: {} - Receiver: {receiver_id}",
+                &*env::predecessor_account_id()
+            );
+            return self
+                .token_mut()
+                .ft_transfer_call(receiver_id, amount, memo, msg);
+        };
+
+        if receiver_id != *CURRENT_ACCOUNT_ID {
+            near_sdk::log!(
```
mitinarseny, line 469: Remove unnecessary logs when merge

poa-token/src/contract.rs
```diff
@@ -108,15 +454,53 @@ impl FungibleTokenCore for Contract {
...
+                "ft_transfer_call destination is not the smart contract address - proceeding with a standard ft_transfer_call for token."
+            );
+            return self
+                .token_mut()
+                .ft_transfer_call(receiver_id, amount, memo, msg);
+        }
+
+        let msg_parts = msg.splitn(3, ':').collect::<Vec<_>>();
```
mitinarseny, line 477: do you really need to allocate a vector here?

tests/old-artifacts/old-poa-token/defuse.wasm
mitinarseny, line 1: remove unnecessary old artifacts

mitinarseny (changes requested):

poa-token/src/contract.rs
BonelessImpl: As discussed, Promise is alright.

poa-token/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 #[cfg(feature = "contract")]
-mod contract;
+pub mod contract;
```
mitinarseny, line 3: Please, remember to add 
```rust
impl ControllerUpgradable for Contract {
    #[only(self, owner)]
    fn upgrade(...) -> Promise { /* ... */}
}
```

poa-token/src/contract.rs
BonelessImpl: I always prefer to have traits impls that exist only to implement an interface take from a local impl. I changed it anyway.

PR #61, opened 2025-04-10
Make storage registration only for owner

BonelessImpl:


poa-token/src/contract.rs
```diff
@@ -142,6 +142,13 @@ impl StorageManagement for Contract {
         account_id: Option<AccountId>,
         registration_only: Option<bool>,
     ) -> StorageBalance {
+        {
```
fusede, line 145:
```suggestion
        #[cfg_attr(feature = "no-registration", only(self, owner))]
```

fusede: or add permalink to near_plugins_macros otherwise

fusede (changes requested):

tests/src/tests/defuse/intents/relayers.rs
```diff
@@ -1,13 +1,19 @@
...
 use crate::{tests::defuse::env::Env, utils::acl::AclExt};
 
 #[tokio::test]
-async fn test_relayer_keys() {
-    let env = Env::builder().deployer_as_super_admin().build().await;
+#[rstest]
+async fn test_relayer_keys(#[values(false, true)] no_registration: bool) {
+    let env = Env::builder()
+        .deployer_as_super_admin()
+        .no_registration(no_registration)
```
fusede, line 14: maybe add `no_registration: bool` matrix only for the tests that are expected to work with transferring PoA FT tokens? Since for this test, it doesn't change anything, so we'll just wait longer for it to finish

Makefile.toml
```diff
@@ -34,7 +36,7 @@ args = [
 ]
 
 [tasks.build-poa-factory]
-dependencies = ["add-cache-dir-tag", "build-poa-token"]
+dependencies = ["add-cache-dir-tag", "build-poa-token", "build-poa-token-no-registration"]
```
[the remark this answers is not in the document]
BonelessImpl: Our tests depend on the no-registration version.

tests/src/tests/defuse/intents/relayers.rs
BonelessImpl: We may reason that this test isn't affected by the no-registration change, but this can be seen as a regression test. Theoretically speaking, there's no way to guarantee that this feature isn't affected due to the general complexity of software. This is why excessively using features is a bad idea, but it multiplies the number of possibilities by 2, instead of just changing some code that's already tested by just rerunning the same tests.

If you insist, we can do this... but it's a bad idea to make assumptions about code coupling and internal relationships.

PR #63, opened 2025-04-18
Remove allowance for relayers

BonelessImpl:


tests/src/tests/defuse/intents/relayers.rs
```diff
@@ -19,34 +24,88 @@ async fn test_relayer_keys(#[values(false, true)] no_registration: bool) {
...
+            env.defuse.id().clone(),
+            new_relayer_secret_key.clone(),
+            &worker,
         )
+        .execute_intents([]) // Empty because it's just to ensure that authorization works/doesn't work
         .await
-        .unwrap_err();
+        .unwrap_err()
+        .to_string()
+        .contains("Failed to query access key")
```
fusede, line 45: test for specific typed errors if available, not for the string representation
https://docs.rs/near-workspaces/latest/near_workspaces/result/struct.ExecutionFinalResult.html#method.into_result

tests/src/tests/defuse/intents/relayers.rs
```diff
@@ -19,34 +24,88 @@ async fn test_relayer_keys(#[values(false, true)] no_registration: bool) {
...
+    );
 
+    // A random, unauthorized user attempts to add a key (no role `Role::RelayerKeysManager`) and fails
+    assert!(
+        env.user2
+            .add_relayer_key(env.defuse.id(), &new_relayer_public_key_near_sdk)
+            .await
+            .unwrap_err()
+            .to_string()
+            .contains("Requires one of these roles:")
```
fusede, line 55: The string representation of this error can be changed in future versions of `near-plugins`, where this error [originates](https://github.com/Near-One/near-plugins/blob/25cb48672fbbd75a9b5e9d1fd0a4f3b3d8acfd35/near-plugins-derive/src/access_controllable.rs#L734) from. Since we don't have a requirement to maintain the same error strings, we shouldn't test for it.
Follow this basic rule: if you don't have a typed error to match against, then do not test for its string representation, as we don't have any control over it. Treat it as a blackbox then

tests/src/tests/defuse/intents/relayers.rs
```diff
@@ -19,34 +24,88 @@ async fn test_relayer_keys(#[values(false, true)] no_registration: bool) {
...
+        .unwrap();
+
+    // Attempt to add a key that already exists
+    assert!(
+        env.user1
+            .add_relayer_key(env.defuse.id(), &new_relayer_public_key_near_sdk)
+            .await
+            .unwrap_err()
+            .to_string()
+            .contains("key already exists")
```
fusede, line 71: Test for specific error variant from `near-workspaces`

tests/src/tests/defuse/intents/relayers.rs
```diff
@@ -19,34 +24,88 @@ async fn test_relayer_keys(#[values(false, true)] no_registration: bool) {
...
+    .unwrap();
+
+    // Attempt to delete the key by an unauthorized user
+    assert!(
+        env.user2
+            .delete_relayer_key(env.defuse.id(), &new_relayer_public_key_near_sdk)
+            .await
+            .unwrap_err()
+            .to_string()
+            .contains("Requires one of these roles:")
```
fusede, line 91: See comment above

tests/src/tests/defuse/intents/relayers.rs
```diff
@@ -19,34 +24,88 @@ async fn test_relayer_keys(#[values(false, true)] no_registration: bool) {
...
+        .unwrap();
+
+    // Delete the same key again, which won't work
+    assert!(
+        env.user1
+            .delete_relayer_key(env.defuse.id(), &new_relayer_public_key_near_sdk)
+            .await
+            .unwrap_err()
+            .to_string()
+            .contains("key not found")
```
fusede, line 107: Test for specific error variant from `near-workspaces`

fusede (changes requested):

tests/src/tests/defuse/intents/relayers.rs
BonelessImpl: The errors are dynamically typed. It's not really possible to get any type information.

tests/src/tests/defuse/intents/relayers.rs
BonelessImpl: Checking against typed errors is a nice. I totally agree with you. However, if we don't have them, then worse things can happen if we ignore them.

Imagine a situation where you have code with two errors, A and B. And the only way to get which error it's, is by getting a string form of the error. Then we create a test:

```
system.cause_error_a().unwrap_err();
system.cause_error_b().unwrap_err();
```

The purpose of these tests is to ensure that the function call `cause_error_a()` causes error A. Now imagine a situation where we refactor the code, and `cause_error_a()` would trigger error B (or any other error, including trivial errors like insufficient gas that have nothing to do with the code logic). Now this breaks many things in the behavior of the code, but the tests won't detect it and the tests will pass!!!! Result: Leaving a hidden bug that we'll never find, unless something big happens, potentially.

To evaluate whether using strings is good or bad, the equation is simple: What's the worst thing that can happen? The tests break and we have to update them? Sure. Who cares? That's much, much cheaper, easier and more manageable than having hidden bugs hidden behind wanting only to test perfect error types. That's how I evaluate this problem.

fusede: These are integration tests where we treat the contract implementation as a blackbox, and only rely on its interface. Our contract's ABI doesn't promise to maintain any error order or exact error strings. All we need is to fail the current receipt if some conditions were or weren't met.

Imagine following situation: you're testing for `add_relayer_keys` and try to add a public key from an account which was not granted `RelayerKeyManager` role, but you didn't attach 1yN to this function call. Currently, `#[access_control_any(roles(...))]` macro adds an ACL check to the beginning of the function, so the FunctionCall will fail with `Insufficient permissions for method add_relayer_key restricted by access control. Requires one of these roles: ...`.

But what if they change the implementation, so this check is added to the bottom of the function? Then you'd see `Requires attached deposit of exactly 1 yoctoNEAR` instead. Nothing changed. The implementation is still not buggy. It just now has different error order / string representation.

I agree that this particular case is very synthetic and it won't happen in reality, but the same can apply to `execute_intents(...)` tests: which error should the contract fail with if you specified already used `nonce`, already expired `deadline` and non-existing public_key on `signer_id`? It's up to [implementation](https://github.com/near/intents/blob/862549febfce15d79c9aabc94e0ee037bcebfd04/core/src/engine/mod.rs#L67-L79). The only thing that implementation MUST do is to fail current receipt.

tests/src/tests/defuse/intents/relayers.rs
fusede: If you want to test for a specific error, then you'll need to change the interface of `add_relayer_key()` to return `Result<(), ExecutionFailure>` instead of `anyhow::Error` to be able to match over it

tests/src/tests/defuse/intents/relayers.rs
TheQuantumPhysicist: This is taken into account. We have two poison pills, and we have to pick one:

1. Accept having hidden bugs that we can't detect after every refactor
2. Accept having to update the tests, if... IF near chganges something, and most likely after a conscious `cargo update` we run.

I'll take option 2 every time, because the correctness of the code behavior is of top priority. The interface perfection isn't important at this test level, integration or not, because we don't have unit tests or anything else to replace it. We're fully relying on these tests to test all internal behavior, including specific errors. If we were testing the internal behavior with unit tests or something else, I would agree with you. But without this, we're 100% naked to hidden bugs.

BonelessImpl: This is taken into account. We have two poison pills, and we have to pick one:

1. Accept having hidden bugs that we can't detect after every refactor
2. Accept having to update the tests, if... IF near changes something, and most likely after a conscious cargo update we run.

I'll take option 2 every time, because the correctness of the code behavior is of top priority. The interface perfection isn't important at this test level, integration or not, because we don't have unit tests or anything else to replace it. We're fully relying on these tests to test all internal behavior, including specific errors. If we were testing the internal behavior with unit tests or something else, I would agree with you. But without this, we're 100% naked to hidden bugs.

tests/src/tests/defuse/intents/relayers.rs
BonelessImpl: Unfortunately, this won't help, because the errors that are typed in ExecutionFailure are near blockchain specific. Everything that comes from the smart contracts (including ACL, plugins, logic, etc) is dynamically typed, even inside ExecutionFailure. After all, the panics that are used are simply strings.

tests/src/tests/defuse/intents/relayers.rs
fusede: The is no hidden bugs unless the you don't have the desired outcome (i.e. state of the contract(s)). Even if there is a hidden bug, which this test case doesn't detect, testing on particular error strings doesn't help to detect it.
The bug would be if this receipt finishes successfully and you'll see a new FunctionCall access key added to the contract, while the signer didn't have permissions to do so.
Again, these are integration tests, and the purpose of these tests is to treat underlying implementation as a blackbox, which relying only on contract's ABI.

fusede (approved):

PR #64, opened 2025-04-18
Add assert_error_contains in a trait to properly check for error string contents

BonelessImpl:
The contents of errors in strings can be checked. If the env var `DEFUSE_SKIP_STRING_ERROR_CHECKS` is defined, only `unwrap_err()` is checked.

### DIFF
--- a/test-utils/src/asserts.rs
+++ b/test-utils/src/asserts.rs
@@ -10,7 +10,7 @@ where
 {
     fn assert_error_contains(&self, to_contain: &str) {
         match self {
-            Ok(_) => panic!("unwrap_err() with Ok() result"),
+            Ok(_) => panic!("Result::unwrap_err() on Result::Ok()"),
             Err(e) => {
                 // Define the env var to check strings in errors
                 let check_string = std::env::var(DISABLE_STRING_CHECKS_ENV_VAR).is_err();
@@ -18,7 +18,7 @@ where
                     let error_string = e.to_string();
                     assert!(
                         e.to_string().contains(to_contain),
-                        "unwrap_err() successful, but the error string does not contain the expected string.\nError string: `{error_string}`\nshould have contained: `{to_contain}`"
+                        "Result::unwrap_err() successful, but the error string does not contain the expected string.\nError string: `{error_string}`\nshould have contained: `{to_contain}`"
                     );
                 } else {
                     eprintln!(
--- a/test-utils/src/asserts.rs
+++ b/test-utils/src/asserts.rs
@@ -1,4 +1,4 @@
-use std::cell::LazyCell;
+use std::{cell::LazyCell, fmt::Display};
 
 const DISABLE_STRING_CHECKS_ENV_VAR: &str = "DEFUSE_SKIP_STRING_ERROR_CHECKS";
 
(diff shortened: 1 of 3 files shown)


test-utils/src/asserts.rs
```diff
@@ -0,0 +1,31 @@
+const DISABLE_STRING_CHECKS_ENV_VAR: &str = "DEFUSE_SKIP_STRING_ERROR_CHECKS";
+
+pub trait ResultAssertsExt {
+    fn assert_error_contains(&self, s: &str);
+}
+
+impl<T, E> ResultAssertsExt for Result<T, E>
+where
+    E: ToString,
```
fusede, line 9:
```suggestion
    E: Display,
```

test-utils/src/asserts.rs
```diff
@@ -0,0 +1,31 @@
...
+{
+    fn assert_error_contains(&self, to_contain: &str) {
+        match self {
+            Ok(_) => panic!("unwrap_err() with Ok() result"),
+            Err(e) => {
+                // Define the env var to check strings in errors
+                let check_string = std::env::var(DISABLE_STRING_CHECKS_ENV_VAR).is_err();
```
fusede, line 16: maybe read it only once and save to static variable?
```rust
thread_local! {
    static DISABLE_STRING_CHECKS: LazyCell<bool> = LazyCell::new(
        || std::env::var(DISABLE_STRING_CHECKS_ENV_VAR).is_ok(),
    );
}
```

test-utils/src/asserts.rs
```diff
@@ -0,0 +1,31 @@
...
+
+pub trait ResultAssertsExt {
+    fn assert_error_contains(&self, s: &str);
+}
+
+impl<T, E> ResultAssertsExt for Result<T, E>
+where
+    E: ToString,
+{
+    fn assert_error_contains(&self, to_contain: &str) {
```
fusede, line 11: Make it accept `impl Display`, so you can `.assert_err_contains(MyCustomError::Variant)` and return `E` for better usage
```suggestion
    #[track_caller]
    fn assert_err_contains(&self, to_contain: impl Display) -> E {
```

fusede (changes requested):

test-utils/src/asserts.rs
BonelessImpl: I'm aware that `ToString` is implemented automatically for anything `Display`, but the other way around isn't necessarily true. Are you sure you want this change?

test-utils/src/asserts.rs
BonelessImpl: It doesn't really matter... it's just tests... but alright.

test-utils/src/asserts.rs
fusede: Yes, `Display` is more idiomatic to use as trait bound, while `ToString` is just a convenient adaptor IMHO.

Some examples from other crates:
* `Result::<T, E>::unwrap()` requires `E: fmt::Debug`, which is also defined in `fmt`
* `anyhow::Context::context<C>(context: C)` requires `C: Display`

fusede (approved):

PR #66, opened 2025-04-22
feat/aml: locking accounts

fusede:
See docs for `AccountForceLocker`:
```rust
#[ext_contract(ext_force_account_locker)]
pub trait AccountForceLocker: AccessControllable {
    /// Returns whether the given`account_id` is locked
    fn is_account_locked(&self, account_id: &AccountId) -> bool;

    /// Locks given `account_id` from modifying its own state, including
    /// token balances.
    /// Returns `false` if the account was already in locked state.
    ///
    /// Attached deposit of 1yN is required for security purposes.
    ///
    /// NOTE: this still allows for force withdrawals/transfers
    fn force_lock_account(&mut self, account_id: AccountId) -> bool;
    /// Unlocks given `account_id`.
    /// Returns `false` if the account wasn't in locked state.
    ///
    /// Attached deposit of 1yN is required for security purposes.
    fn force_unlock_account(&mut self, account_id: &AccountId) -> bool;
}
```

## Summary by CodeRabbit

- **New Features**
  - Introduced account locking and unlocking functionality, allowing accounts to be locked to prevent outgoing transfers and state modifications, with force unlock/lock operations available to privileged roles.
  - Added forced multi-token transfer and withdrawal methods, enabling privileged roles to bypass account locks for specific operations.
  - Added support for custom serialization and versioning of account data, ensuring backward compatibility and safe upgrades.
  - Introduced utility crates and traits for advanced serialization, deserialization, and I/O operations.

- **Enhancements**
  - Improved error messages with more contextual information for account and key operations.
  - Added explicit lock state checks and error handling throughout account and token management flows.
  - Expanded role-based access control with new roles for unrestricted account locking/unlocking and forced token operations.
  - Improved test coverage with comprehensive tests for account locking, unlocking, and key management.

- **Bug Fixes**
  - Ensured deposits are allowed to locked accounts, while restricting outgoing transfers and key modifications as intended.

- **Refactor**
  - Centralized account state management and error handling.
  - Updated method signatures to return detailed error results instead of simple booleans.
  - Standardized naming for workspace dependencies and crates.

- **Documentation**
  - Enhanced documentation and type clarity for public interfaces and new features.

- **Chores**
  - Added new utility crates and updated workspace configuration for consistency and maintainability.
  - Updated test and development dependencies for improved reliability and reproducibility.


near-utils/src/panic_on_clone.rs
```diff
@@ -0,0 +1,57 @@
...
+    use core::ptr;
+
+    use super::*;
+
+    #[test]
+    fn from_ref() {
+        let value = "example".to_string();
+        let poc = PanicOnClone::from_ref(&value);
+        assert!(ptr::eq(&**poc, &value));
+        assert_eq!(&**poc, &value);
```
BonelessImpl, line 55: No test for panic behavior?

near-utils/src/lock.rs
```diff
@@ -35,81 +41,176 @@ impl<T> Lock<T> {
         Self::new(value, true)
     }
 
+    #[inline]
+    pub const fn set_locked(&mut self, locked: bool) -> &mut Self {
```
BonelessImpl, line 45: I would call this `with_locked`

near-utils/src/lock.rs
BonelessImpl, line 1: I think this struct needs some unit tests... too many special cases.

core/src/error.rs
```diff
@@ -7,8 +8,11 @@ pub type Result<T, E = DefuseError> = ::core::result::Result<T, E>;
 
 #[derive(Debug, ThisError, FunctionError)]
 pub enum DefuseError {
-    #[error("account not found")]
-    AccountNotFound,
+    #[error("account '{0}' not found")]
+    AccountNotFound(AccountId),
+
+    #[error("account '{0}' is locked")]
+    AccountLocked(AccountId),
```
BonelessImpl, line 15: If I understand correctly what "locked" means here, I think the proper language is "frozen"... we "freeze" bank accounts.

near-utils/src/lock.rs
BonelessImpl: set_ usually doesn't return self. That's why. But I don't mind.

tests/src/tests/defuse/accounts.rs
```diff
@@ -100,3 +153,360 @@ impl AccountManagerExt for near_workspaces::Contract {
...
+        &self,
+        defuse_contract_id: &AccountId,
+        account_id: &AccountIdRef,
+    ) -> anyhow::Result<bool>;
+    async fn force_unlock_account(&self, account_id: &AccountIdRef) -> anyhow::Result<bool>;
+    async fn defuse_force_unlock_account(
+        &self,
+        defuse_contract_id: &AccountId,
+        account_id: &AccountIdRef,
+    ) -> anyhow::Result<bool>;
```
BonelessImpl, line 176: I really don't think it's necessary to duplicate these methods... just one of them is enough, and specifying the smart contracts in the tests is alright.

defuse/src/tokens/nep245.rs
```diff
@@ -48,3 +48,48 @@ pub trait MultiTokenForceWithdrawer: MultiTokenWithdrawer + AccessControllable {
         msg: Option<String>,
     ) -> PromiseOrValue<Vec<U128>>;
 }
+
+#[ext_contract(ext_mt_force_core)]
+pub trait MultiTokenForceCore: MultiTokenCore + AccessControllable {
```
BonelessImpl, line 53: Perhaps the super trait should be `AccountForceLocker`, not `AccessControllable`... what's the point of forcing if the contract doesn't have the locking functionality?

defuse/src/accounts.rs
```diff
@@ -31,3 +32,23 @@ pub trait AccountManager {
     /// NOTE: MUST attach 1 yⓃ for security purposes.
     fn invalidate_nonces(&mut self, nonces: Vec<AsBase64<Nonce>>);
 }
+
+#[ext_contract(ext_force_account_locker)]
+pub trait AccountForceLocker: AccessControllable {
```
BonelessImpl, line 37: Besides using the word freeze instead of lock, I think this trait's name should be something like AccountLockable. If we agree on using freeze, then it's AccountFreezable.

near-utils/src/panic_on_clone.rs
```diff
@@ -0,0 +1,57 @@
...
+use near_sdk::{env, near};
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[autoimpl(Deref using self.0)]
+#[autoimpl(DerefMut using self.0)]
+#[autoimpl(AsRef using self.0)]
+#[autoimpl(AsMut using self.0)]
+#[near(serializers = [borsh])]
+#[repr(transparent)] // needed for `transmute()` below
+pub struct PanicOnClone<T: ?Sized>(T);
```
BonelessImpl, line 11: Perhaps write some docs that explain the use-case of this struct, i.e., what problem is solves with a little example.

defuse/src/contract/accounts/account.rs
```diff
@@ -141,3 +162,90 @@ enum AccountPrefix {
     PublicKeys,
     State,
 }
+
+/// Versioned [Account] state for de/serialization
+#[derive(Debug)]
+#[near(serializers = [borsh])]
+enum VersionedAccountEntry<'a> {
+    V1(Cow<'a, PanicOnClone<Account>>),
+    V2(Cow<'a, PanicOnClone<Lock<Account>>>),
```
BonelessImpl, line 171: Why is it needed to store the state of locking with the account object? This seems to have complicated things a lot... isn't easier to simply add a map of locked accounts, and simply gate ops that we want to lock?

If I have a choice between an upgrade that can happen once, where we add a HashSet/HashMap to account for locking, and having such a complex dynamic system, I'd always choose the former... maybe I'm missing something?

defuse/src/contract/tokens/nep245/resolver.rs
```diff
@@ -51,8 +51,25 @@ impl MultiTokenResolver for Contract {
...
+                // for this account. If we allow refunds to happen, then it would lead to
+                // `mt_transfer` events emitted with `old_owner_id` being the locked account,
+                // which we want to avoid.
+                //
+                // So, since we allow locked accounts to receive incoming transfers, it means
+                // that `mt_transfer_call()` to locked recepients will always result in full
+                // transfer without any refunds, even if the receiver returned non-zero refund
+                // amounts from `mt_on_transfer`.
+                .and_then(Lock::as_unlocked_mut)
+            else {
```
BonelessImpl, line 70: I think it might be OK to refund to a locked/frozen account. This is because the described race condition is very unlikely to happy. It's more important to maintain the sanity of the behavior than worry about "loss of funds"... after all, we have the ultimate control on the funds.

BonelessImpl (commented):
Done with the first review round.

near-utils/src/panic_on_clone.rs
BonelessImpl: /// This struct is used as a tool to make it possible to derive borsh serialization in such a way where serialization can take over a reference and include it as if it's owned by the struct/enum being serialized. The reference can be taken in the function `from_ref()`.

borsh-utils/src/lib.rs
```diff
@@ -1,2 +1,3 @@
+pub mod r#as;
```
BonelessImpl, line 1: Can we call this something that doesn't use a literal prefix? Maybe ser_as?

io-utils/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        TeeReader {
+            reader: self,
+            writer,
+        }
+    }
+}
+impl<R> ReadExt for R where R: Read {}
+
+pub struct TeeReader<R, W> {
```
BonelessImpl, line 17: /// A reader that writes everything it reads to the specified writer. Can bee seen as a passthrough. 
/// The struct itself does not do any kind of caching. Any apparent caching is possibly done by the inner read and writer.

defuse/src/tokens/nep245.rs
```diff
@@ -48,3 +48,48 @@ pub trait MultiTokenForceWithdrawer: MultiTokenWithdrawer + AccessControllable {
         msg: Option<String>,
     ) -> PromiseOrValue<Vec<U128>>;
 }
+
```
BonelessImpl, line 51: /// `MultiTokenCore` functions, but a version that bypasses the locked/frozen state of an account.

BonelessImpl: /// All functions must be `#[private]` or only authorized callers

defuse/src/accounts.rs
```diff
@@ -31,3 +32,23 @@ pub trait AccountManager {
...
+    fn is_account_locked(&self, account_id: &AccountId) -> bool;
+
+    /// Locks given `account_id` from modifying its own state, including
+    /// token balances.
+    /// Returns `false` if the account was already in locked state.
+    ///
+    /// Attached deposit of 1yN is required for security purposes.
+    ///
+    /// NOTE: this still allows for force withdrawals/transfers
+    fn force_lock_account(&mut self, account_id: AccountId) -> bool;
```
BonelessImpl, line 48: Instead of returning a bool here (and in unlock), we should probably just panic with a message "already locked".

defuse/src/accounts.rs
```diff
@@ -21,7 +22,7 @@ pub trait AccountManager {
     /// i.e. this key can't be used to make any actions unless it's re-created.
     ///
     /// NOTE: MUST attach 1 yⓃ for security purposes.
-    fn remove_public_key(&mut self, public_key: &PublicKey);
+    fn remove_public_key(&mut self, public_key: PublicKey);
```
BonelessImpl, line 25: Don't we usually prefer passing a reference?

near-utils/src/lock.rs
```diff
@@ -35,81 +41,176 @@ impl<T> Lock<T> {
...
         }
-        Some(&self.value)
+        Some(self.as_inner_unchecked())
     }
 
+    #[must_use]
     #[inline]
-    pub fn as_unlocked_mut(&mut self) -> Option<&mut T> {
-        if self.locked {
+    pub const fn as_unlocked_mut(&mut self) -> Option<&mut T> {
```
BonelessImpl, line 135: Perhaps a better name for all "as_unlocked*" functions is "get*".

near-utils/src/lock.rs
```diff
@@ -35,81 +41,176 @@ impl<T> Lock<T> {
...
         }
-        Some(&mut self.value)
+        Some(self.as_inner_unchecked_mut())
     }
 
+    #[must_use]
     #[inline]
-    pub fn unlock(&mut self) -> Option<&mut T> {
-        if !self.locked {
+    pub const fn as_unlocked_or_mut(&mut self, force: bool) -> Option<&mut T> {
```
BonelessImpl, line 144: `get_force_mut`?

near-utils/src/lock.rs
```diff
@@ -35,81 +41,176 @@ impl<T> Lock<T> {
...
+
+    #[must_use]
     #[inline]
     pub const fn is_locked(&self) -> bool {
         self.locked
     }
 
+    #[must_use]
     #[inline]
     pub const fn as_locked(&self) -> Option<&T> {
```
BonelessImpl, line 73: Do we really need this function? What's the purpose of a function that gets while asserting the lock is activated? It should be either "I don't care about the lock" or "Get only if unlocked". This feels more like technical jargon to me.

defuse/src/tokens/nep245.rs
mitinarseny: `MultiTokenForceCore` is intended to be used by a permissioned account to be able to forcibly transfer tokens from one account to another. As an addition it also bypasses locks.

defuse/src/contract/tokens/nep245/resolver.rs
mitinarseny: Refunds **to** locked accounts are OK and allowed here. This implementation restricts refunds **from** locked accounts, since it would result in `mt_transfer` event  emitted with `old_owner_id` being the locked account, which shouldn't be the case as soon as we lock the account

near-utils/src/lock.rs
BonelessImpl: Can we remove these now to make the API simpler to understand? If we need these functions in the future, we can assess whether we need them and add them if we do.

defuse/src/contract/accounts/account.rs
BonelessImpl: We have discussed this formerly and I expressed my extreme disagreement with this approach due to its complexity. You are welcome to veto my decision and do what you wish. I will resolve this comment.

tests/src/tests/defuse/env.rs
```diff
@@ -228,17 +254,33 @@ impl EnvBuilder {
...
+
+        for role in self.self_as_grantee {
             self.roles
-                .super_admins
-                .insert(format!("defuse.{}", root.id()).parse().unwrap());
+                .grantees
+                .entry(role)
+                .or_default()
+                .insert(defuse_contract_id.clone());
+        }
```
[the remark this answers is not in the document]
BonelessImpl: I misunderstood how this is used. Never mind.

defuse/src/contract/accounts/account.rs
```diff
@@ -141,3 +166,141 @@ enum AccountPrefix {
...
 }
+
+/// This is a magic number that is used to differentiate between
+/// borsh-serialized representations of legacy and versioned [`Account`]s:
+/// * versioned [`Account`]s always start with this prefix
+/// * legacy [`Account`] starts with other 4 bytes
+///
+/// This is safe to assume that legacy [`Account`] doesn't start with
+/// this prefix, since the first 4 bytes in legacy [`Account`] were used
+/// to detote the length of `prefix: Box<[u8]>` in [`LookupMap`] for
```
BonelessImpl, line 177: "denote"?

defuse/src/contract/accounts/account.rs
```diff
@@ -141,3 +166,141 @@ enum AccountPrefix {
...
+        } else {
+            Account::deserialize_reader(
+                // prepend already consumed part of the reader
+                &mut buf.chain(reader),
+            )
+            .map(Into::into)
+        }
+        .map(Into::into)
+    }
+}
```
BonelessImpl, line 240: Since we're defining the serialization manually anyway, isn't it better to define the serialization schema without PanicOnClone? We're crossing the barrier of "everything is just derived" to "we have to customize it"... let's go all the way and serialize the enum manually, and deserialize it manually as well. We're already doing that. It would be really awesome of we could throw out PanicOnClone.

defuse/src/contract/accounts/account.rs
```diff
@@ -141,3 +166,141 @@ enum AccountPrefix {
...
+    use near_sdk::borsh;
+    use rstest::rstest;
+
+    use crate::contract::{Prefix, accounts::AccountsPrefix};
+
+    use super::*;
+
+    #[rstest]
+    #[test]
+    fn legacy_upgrade(random_seed: Seed) {
```
BonelessImpl, line 269: This is an upgrade test. For such complex serialization schemes, we need fixed value tests too (serialized hex values, back and forth in both cases). This helps in avoiding bugs like the Nest thing, where we need that the complicated scheme ended with something we understand.

To illustrate what I see here: Imagine a function f(x) and its inverse g(x). The test here is kind of testing `f(g(x)) = 1`. But we need a test for what f(x) alone does, and what g(x) alone does too, because there are trivial cases where `f(g(x)) = 1` is always true, but both f and g are broken. 

IMHO, this format of testing must exist in any non-standard serialization plan.

fusede:
Closing in favor of #106

PR #69, opened 2025-04-23
feat: let users to disable auth by PREDECESSOR_ID

fusede:


fusede:
Closed in favor of #112

PR #70, opened 2025-04-24
fix: NFT owner_id is not contract_id, but real owner id

fusede:


defuse/src/contract/tokens/mod.rs
```diff
@@ -36,10 +36,17 @@ impl Contract {
...
+                .state
                 .total_supplies
                 .add(token_id.clone(), amount)
                 .ok_or(DefuseError::BalanceOverflow)?;
+            match token_id {
+                TokenId::Nep171(contract_id, token_id) if total_supply > 1 => {
+                    return Err(DefuseError::NftAlreadyDeposited(contract_id, token_id));
+                }
+                _ => {}
+            }
```
BonelessImpl, line 49: I don't believe we have (good) tests for NFTs... we should probably do that.

defuse/src/contract/tokens/mod.rs
```diff
@@ -36,10 +36,17 @@ impl Contract {
...
+            let total_supply = self
+                .state
                 .total_supplies
                 .add(token_id.clone(), amount)
                 .ok_or(DefuseError::BalanceOverflow)?;
+            match token_id {
+                TokenId::Nep171(contract_id, token_id) if total_supply > 1 => {
+                    return Err(DefuseError::NftAlreadyDeposited(contract_id, token_id));
+                }
+                _ => {}
```
BonelessImpl, line 48: I would change this to manually handling each variant. Catch-all arms have the risk that any newly added arm in the future won't trigger an error, leading to possible bugs.

BonelessImpl (approved):

defuse/src/contract/tokens/mod.rs
fusede: Ok, created a task in backlog

PR #72, opened 2025-05-05
Fix deposit mint event for NEP-245

BonelessImpl:


defuse/src/contract/tokens/mod.rs
```diff
@@ -53,7 +53,19 @@ impl Contract {
...
 
-        MtEvent::MtMint([mint_event].as_slice().into()).emit();
+        // Emit the event only if the accumulative amount of all tokens is > 0
+        {
+            let accumulative_amounts = mint_event
+                .amounts
+                .iter()
+                .try_fold(0u128, |accum, curr| accum.checked_add(curr.0));
+
+            if let Some(v) = accumulative_amounts {
```
BonelessImpl, line 63: There's a question here on what to do if `accumulative_amounts` is `None`. Using `try_fold` is done out of caution, but there's no reason it can fail if everything before has failed. If `try_fold` fails, then some invariant is broken.

So, we either dismiss this and forget about it and hope to catch it if it ever happens (i.e., there's an overflow that does nothing except break the event emission); or, we panic or emit logs to find this issue in the future.

fusede (approved):

PR #74, opened 2025-05-06
Add more information to SimulationOutput

BonelessImpl:


BonelessImpl:
I could not find a way to add ft_withdraw information. The issue is that it might require interface changes to both the Engine, which works with and without a simulation.

PR #75, opened 2025-05-07
Add NFT tests for the verifier contract

BonelessImpl:


tests/src/utils/nft.rs
```diff
@@ -1,9 +1,23 @@
...
+    env!("CARGO_MANIFEST_DIR"),
+    "/contracts/non-fungible-token.wasm"
+));
+
 pub trait NftExt {
+    async fn deploy_vanilla_nft_issuer(
+        &self,
+        token_name: &str,
+        reference: Option<String>,
+        reference_hash: Option<Base64VecU8>,
```
fusede, line 18: I'd suggest to move wrapping into `Base64VecU8` inside the implementation, since it's only needed for serialization purposes. And take `Option<near_sdk::CryptoHash>` as parameter, since the [standard](https://nomicon.io/Standards/Tokens/FungibleToken/Metadata#an-implementing-contract-may-include-the-following-fields-on-chain) says:
> reference_hash: the base64-encoded sha256 hash of the JSON file contained in the reference field. This is to guard against off-chain tampering.

fusede: Oh, of maybe reuse already existing [NFTContractMetadata](near_contract_standards::non_fungible_token::metadata::NFTContractMetadata). It still requires `Option<Base64VecU8>`, but might be better to use already existing definition.

tests/src/utils/nft.rs
```diff
@@ -31,6 +53,48 @@ pub trait NftExt {
...
+    ) -> anyhow::Result<Contract> {
+        let contract = self
+            .deploy_contract(token_name, NON_FUNGIBLE_TOKEN_WASM)
+            .await?;
+
+        let mut args = json!({
+            "owner_id": self.id(),
+            "metadata": {
```
fusede, line 68: Just serialize `NFTContractMetadata` above

fusede (changes requested):

fusede (approved):

PR #77, opened 2025-05-09
Add more information about withdrawals in MtBurn

BonelessImpl:


BonelessImpl (commented):
This is a prototype of the solution. This probably needs to be discussed.

fusede (approved):

PR #79, opened 2025-05-16
Post events at the execution level and horde them in the inspector

BonelessImpl:


defuse/src/contract/intents/simulate.rs
```diff
@@ -1,30 +1,28 @@
...
 pub struct SimulateInspector {
     pub intents_executed: Vec<IntentEvent<AccountEvent<'static, ()>>>,
     pub min_deadline: Deadline,
     pub balance_diff: HashMap<AccountId, TokenDeltas>,
+    #[allow(dead_code)] // FIXME: remove
     pub wnear_id: AccountId,
     pub ft_withdrawals: Option<Vec<FtWithdraw>>,
     pub nft_withdrawals: Option<Vec<NftWithdraw>>,
     pub mt_withdrawals: Option<Vec<MtWithdraw>>,
+    pub events_emitted: Vec<String>,
```
fusede, line 25: Make it more convenient for callers to avoid parsing strings as JSON:
```suggestion
    pub events: Vec<serde_json::Value>,
```

core/src/engine/inspector.rs
```diff
@@ -1,68 +1,13 @@
...
-        intent_hash: CryptoHash,
-    ) -> Result<()>;
-
-    fn on_ft_withdraw(
-        &mut self,
-        owner_id: &AccountIdRef,
-        ft_withdraw: &FtWithdraw,
-        intent_hash: CryptoHash,
-    ) -> Result<()>;
+    fn emit_event(&mut self, event: DefuseEvent<'_>);
```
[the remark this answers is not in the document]
fusede: This way it will only be possible to emit `dip4` events, but not `nep245` and others...
Can you make it accept all possible events that we might have in some structured way? E.g. by accepting `Event` enum as parameter:
```rust
#[serde(untagged)]
 pub enum Event<'a> {
    Dip4(DefuseEvent<'a>),
    Nep245(MtEvent<'a>),
    // ...
}

impl Event<'_> {
    pub fn emit(&self) {
        match self {
            Self::Dip4(event) => event.emit(),
            Self::Nep245(event) => event.emit(),
        }
    }
}
```

The complex part will be to re-write `Contract::deposit()` and `Contract::withdraw` methods, so that they become `Inspector`-aware, or, in general, intent-aware... Or maybe you have other suggestions?

defuse/src/contract/intents/simulate.rs
fusede: Or it would be even better if we don't bind to JSON specifically, but use strongly-typed `Event` struct/enum...

BonelessImpl: Push event objects presents a challenge because the values are mostly references. In fact, I changed some Cow::Borrow to Owned to make this work. If we get rid of references, then this is easily doable.

defuse/src/contract/intents/simulate.rs
```diff
@@ -1,30 +1,28 @@
...
 
 pub struct SimulateInspector {
     pub intents_executed: Vec<IntentEvent<AccountEvent<'static, ()>>>,
     pub min_deadline: Deadline,
     pub balance_diff: HashMap<AccountId, TokenDeltas>,
+    #[allow(dead_code)] // FIXME: remove
     pub wnear_id: AccountId,
     pub ft_withdrawals: Option<Vec<FtWithdraw>>,
     pub nft_withdrawals: Option<Vec<NftWithdraw>>,
     pub mt_withdrawals: Option<Vec<MtWithdraw>>,
```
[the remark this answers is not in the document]
BonelessImpl: Sure. I kept them just to see what plans you have. I'll get rid of them.

defuse/src/contract/intents/simulate.rs
```diff
@@ -17,34 +13,25 @@ pub struct SimulateInspector {
     pub intents_executed: Vec<IntentEvent<AccountEvent<'static, ()>>>,
     pub min_deadline: Deadline,
     pub balance_diff: HashMap<AccountId, TokenDeltas>,
-    #[allow(dead_code)] // FIXME: remove
-    pub wnear_id: AccountId,
-    pub ft_withdrawals: Option<Vec<FtWithdraw>>,
-    pub nft_withdrawals: Option<Vec<NftWithdraw>>,
-    pub mt_withdrawals: Option<Vec<MtWithdraw>>,
-    pub events_emitted: Vec<String>,
+    pub events_emitted: Vec<serde_json::Value>,
```
fusede, line 16:
```suggestion
    pub events: Vec<serde_json::Value>,
```

PR #86, opened 2025-05-21
feat: TON signatures

fusede:


core/src/payload/ton_connect.rs
```diff
@@ -0,0 +1,36 @@
...
+impl<T> ExtractDefusePayload<T> for TonConnectPayload
+where
+    T: DeserializeOwned,
+{
+    type Error = serde_json::Error;
+
+    fn extract_defuse_payload(self) -> Result<DefusePayload<T>, Self::Error> {
+        if self.timestamp > defuse_near_utils::time::now() {
+            return Err(Error::custom("timestamp has not come yet"));
+        }
```
BonelessImpl, line 30: Why does the extraction of the payload depend on the time? We should document this in the code somewhere.

near-utils/src/digest.rs
BonelessImpl, line 1: We need a test for sha256 output, at least one, to ensure that the output we're getting from a certain format is what we expect. This is similar to the logic of why we test serialization against fixed values.

ton-connect/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+    },
+    Binary {
+        #[serde_as(as = "Base64")]
+        bytes: Vec<u8>,
+    },
+    Cell {
+        schema_crc: u32,
+        #[serde_as(as = "AsBoC<Base64>")]
+        cell: Cell,
+    },
```
BonelessImpl, line 131: Can we add some documentation on these variants and why each exists and what problem it solves?

ton-connect/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+                    _ => unreachable!(),
+                };
+                Ok(env::sha256_array(
+                    &[
+                        [0xff, 0xff].as_slice(),
+                        b"ton-connect/sign-data/",
+                        &self.address.workchain_id.to_be_bytes(),
+                        &self.address.address,
+                        &u32::try_from(self.domain.len())
+                            .map_err(|_| Error::custom("domain: too long"))?
```
BonelessImpl, line 63: I would call this overflow... not too long.

ton-connect/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+                            .to_be_bytes(),
+                        self.domain.as_bytes(),
+                        &timestamp.to_be_bytes(),
+                        payload_prefix,
+                        &u32::try_from(payload.len())
+                            .map_err(|_| Error::custom("payload: too long"))?
```
BonelessImpl, line 69: I would call this overflow... not too long.

ton-connect/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+                payload: TonConnectPayloadSchema::Text {
+                    text: "Hello, TON!".repeat(100),
+                },
+            },
+            public_key: hex!("22e795a07e832fc9084ca35a488a711f1dbedef637d4e886a6997d93ee2c2e37"),
+            signature: hex!(
+                "7bc628f6d634ab6ddaf10463742b13f0ede3cb828737d9ce1962cc808fbfe7035e77c1a3d0b682acf02d645cc1a244992b276552c0e1c57d30b03c2820d73d01"
+            ),
+        });
+    }
```
BonelessImpl, line 197: It's also a good idea to test tampering with data. Meaning: Tampering with data will result in failed verification of the signature. I personally have always even tested tampering with individual field. I used an rng to create a mutation in the data, and would always test with and without the mutation for all fields.

BonelessImpl (commented):
First review round complete. Generally speaking the code looks fine, up to the degree of my understanding of the matter. There is, however, a lack of testing and documentation. I know we've always cared less about that because the code was new, but it's a good idea to start with these things.

ton-connect/src/lib.rs
fusede: Variants for this struct come from the TON Connect standard as described in the comment above:
```rust
/// See <https://docs.tonconsole.com/academy/sign-data#choosing-the-right-format>
```

BonelessImpl (approved):
Let's have a discussion about the implementation details today. Besides that, I think the PR looks good.

PR #92, opened 2025-06-01
feat: Add NEP245 Multi Token Metadata Extension

kuksag:
### Overiew
This PR introduces interface required to implement [NEP245 metadata extension](https://nomicon.io/Standards/Tokens/MultiToken/Metadata)

### `Token` struct extension
In the specification it's stated:
>  A new attribute MUST be added to each Token struct: 
> ```diff
> type Token = {
>   token_id: string,
> +  token_metadata?: MTTokenMetadata,
> +  base_metadata_id: string,
> }
> ```
Which I haven't done on purpose.

Technically, it can be implemented using rust compile features to enable/disable code in case of implementing an extension.
But that's actually a breaking change with the code that has already implemented "core" standard.

On the other hand, having a `token_id` is sufficient to get related `token_metadata` and `base_metadata_id` (which forms `MTTokenMetadataAll` combined together) by calling `mt_metadata_token_all`.

## Summary by CodeRabbit

- **New Features**
  - Introduced support for NEP-245 multi-token metadata, including contract-level, base token, and individual token metadata structures.
  - Added public access to the metadata module for managing and retrieving multi-token metadata.
  - Enabled paginated listing and retrieval of metadata for multiple tokens.
  - Improved serialization and deserialization of UTC datetime fields in metadata.

- **Chores**
  - Updated and refined dependencies to support new metadata features and serialization requirements.


nep245/src/metadata.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub expires_at: Option<u64>,
+
+    /// Unix epoch in milliseconds when this token starts being valid, or `None`
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub starts_at: Option<u64>,
+
+    /// Unix epoch in milliseconds when this token metadata was last updated, or `None`
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub updated_at: Option<u64>,
```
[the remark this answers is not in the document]
mitinarseny: Hey @kuksag, thanks again for your contribution!
@Sam is most likely talking about changing `u64` to `chrono::Datetime<Utc>`, so that it's easier to work with. In order to deserialize it as UNIX timestamp (in milliseconds), we suggest using `#[serde_as(as = "TimestampMilliSeconds")]` annotation similar to [this one](https://github.com/near/intents/blob/623da1d0a53ce77c8a28b258676647435a5a2777/ton-connect/src/lib.rs#L42-L43)

kuksag:
I haven't found borsh implementation for `Datetime<Utc>`, so I write a small wrapper. Let me know if that works for you.
Another option is to lift off `#[near(serializers = [borsh])]` requirements, but then anyway it will require every user to come up with some solution to store that data

nep245/src/metadata.rs
```diff
@@ -0,0 +1,199 @@
...
+    #[test]
+    fn test_datetime_utc_wrapper_borsh() {
+        let timestamp = DateTime::from_timestamp(1747772412, 0).unwrap();
+        let wrapped = DatetimeUtcWrapper(timestamp);
+        let mut buffer = Vec::<u8>::new();
+        BorshSerialize::serialize(&wrapped, &mut buffer).unwrap();
+        let actual_wrapped: DatetimeUtcWrapper =
+            BorshDeserialize::deserialize(&mut buffer.as_slice()).unwrap();
+        assert_eq!(actual_wrapped.0, wrapped.0);
+    }
```
BonelessImpl, line 198: Please change the test to this:

```
    #[test]
    fn test_datetime_utc_wrapper_borsh() {
        let timestamp = DateTime::from_timestamp(1747772412, 0).unwrap();
        let wrapped = DatetimeUtcWrapper(timestamp);
        let encoded = borsh::to_vec(&wrapped).unwrap();
        assert_eq!(encoded, Vec::from_hex("60905aef96010000").unwrap());
        let actual_wrapped: DatetimeUtcWrapper = borsh::from_slice(encoded.as_slice()).unwrap();
        assert_eq!(actual_wrapped.0, wrapped.0);
    }
```

and add `hex` to dev-dependencies. There's nothing wrong with what you did, but we need also a fixed value for the test as a reference.

BonelessImpl: Btw, I tried to push these changes myself, but apparently I can't do it due to permissions.

BonelessImpl (approved):
Besides the minor test issue, the interface looks alright. No implementation seems to be available.

kuksag:
nit: perhaps implementation of borsh de-serialization should be moved to borsh-utils

mitinarseny:
Closing due to inactivity

PR #94, opened 2025-06-03
Limit token id length

BonelessImpl:


test-utils/src/account_id.rs
```diff
@@ -0,0 +1,36 @@
...
+            .map(|_| {
+                let c = u.int_in_range(0..=35)?;
+                Ok(match c {
+                    0..=25 => (b'a' + c) as char,
+                    26..=35 => (b'0' + (c - 26)) as char,
+                    _ => unreachable!(),
+                })
+            })
+            .collect::<arbitrary::Result<_>>()?;
+        let s = s + ".near";
```
fusede, line 19: `.near` is just one example of [TLA](https://nomicon.io/DataStructures/Account#top-level-accounts)

fusede (changes requested):

core/src/deadline.rs
```diff
@@ -40,7 +40,7 @@ impl Deadline {
 
     #[must_use]
     #[inline]
-    pub fn into_timestamp(self) -> DateTime<Utc> {
+    pub const fn into_timestamp(self) -> DateTime<Utc> {
```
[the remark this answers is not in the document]
BonelessImpl: This is suggested by clippy.

core/src/engine/state/cached.rs
```diff
@@ -216,15 +216,19 @@ where
             return Err(DefuseError::InvalidIntent);
         }
 
+        let token_ids = std::iter::repeat(withdraw.token.clone())
```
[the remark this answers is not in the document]
BonelessImpl: I'm sorry, what does that mean? Is this comment about `std::`? Killing all namespaces is considered a bad practice, because it can lead to conflicts. Especially for common words like `iter`. But generally I don't really care.

test-utils/src/account_id.rs
```diff
@@ -0,0 +1,36 @@
...
+        s.parse().map_err(|_| arbitrary::Error::IncorrectFormat)
+    } else {
+        // Explicit numeric account id
+        let len = 64;
+        let s: String = (0..len)
```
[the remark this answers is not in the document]
BonelessImpl: `<[u8; 32]>::arbitrary()` does not seem to exist.

https://docs.rs/arbitrary/latest/arbitrary/trait.Arbitrary.html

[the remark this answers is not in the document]
BonelessImpl: I invented a solution of my own.

test-utils/src/account_id.rs
```diff
@@ -0,0 +1,36 @@
...
+                Ok(match c {
+                    0..=25 => (b'a' + c) as char,
+                    26..=35 => (b'0' + (c - 26)) as char,
+                    _ => unreachable!(),
+                })
+            })
+            .collect::<arbitrary::Result<_>>()?;
+        let s = s + ".near";
+        s.parse().map_err(|_| arbitrary::Error::IncorrectFormat)
+    } else {
```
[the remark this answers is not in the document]
BonelessImpl: Done

core/src/engine/state/cached.rs
```diff
@@ -216,15 +216,19 @@ where
...
+            .zip(withdraw.token_ids.iter().cloned())
+            .map(|(token, token_id)| Nep245TokenId::new(token, token_id).map(Into::into))
+            .collect::<Result<Vec<_>, _>>()?;
```
fusede, line 222:
```suggestion
            .map(|(token, token_id)| Nep245TokenId::new(token, token_id))
            .collect::<Result<Vec<_>, _>>()?;
```

core/src/engine/state/cached.rs
```diff
@@ -216,15 +216,19 @@ where
...
+
         self.internal_sub_balance(
             owner_id,
-            std::iter::repeat(withdraw.token.clone())
-                .zip(withdraw.token_ids.iter().cloned())
-                .map(|(token, token_id)| TokenId::Nep245(token, token_id))
+            token_ids
+                .into_iter()
```
fusede, line 227:
```suggestion
                .into_iter()
                .map(Into::into)
```

core/src/engine/state/cached.rs
```diff
@@ -216,15 +216,19 @@ where
...
                 .zip(withdraw.amounts.iter().map(|a| a.0))
                 .chain(withdraw.storage_deposit.map(|amount| {
                     (
-                        TokenId::Nep141(self.wnear_id().into_owned()),
+                        Nep141TokenId::new(self.wnear_id().into_owned()).into(),
```
fusede, line 231: Does it  make sense to add it as default implementation on `StateView`?
```rust
pub trait StateView {
    pub fn wnear_id(&self) -> Cow<'_, AccountIdRef>;
    pub fn wnear_token_id(&self) -> TokenId {
        Nep141TokenId::new(self.wnear_id().into_owned()).into()
    }
    // ...
}
```

core/src/error.rs
```diff
@@ -37,8 +36,8 @@ pub enum DefuseError {
     #[error("JSON: {0}")]
     JSON(#[from] serde_json::Error),
 
-    #[error("NFT '{}' was already deposited", TokenId::Nep171(.0.clone(), .1.clone()))]
-    NftAlreadyDeposited(AccountId, defuse_nep245::TokenId),
+    #[error("NFT '{0}' is already deposited")]
```
fusede, line 39: This will not include `nep171:` prefix, so maybe wrap in `TokenId`?
```suggestion
    #[error("NFT '{}' is already deposited", TokenId::Nep171(.0.clone()))]
```

core/src/intents/token_diff.rs
```diff
@@ -77,7 +75,8 @@ impl ExecutableIntent for TokenDiff {
             // take fees only from negative deltas (i.e. token_in)
             if delta < 0 {
                 let amount = delta.unsigned_abs();
-                let fee = Self::token_fee(&token_id, amount, protocol_fee).fee_ceil(amount);
+                let fee =
+                    Self::token_fee((&token_id).into(), amount, protocol_fee).fee_ceil(amount);
```
fusede, line 79: Maybe then iterate over references above?
```rust
for (token_id, delta) in &self.diff {
```

core/src/intents/token_diff.rs
```diff
@@ -219,12 +218,13 @@ impl TokenDiff {
     }
 
     #[inline]
-    pub const fn token_fee(token_id: &TokenId, amount: u128, fee: Pips) -> Pips {
+    pub const fn token_fee(token_id: TokenIdType, amount: u128, fee: Pips) -> Pips {
         match token_id {
```
fusede, line 222:
```suggestion
    pub const fn token_fee(token_id: impl Into<TokenIdType>, amount: u128, fee: Pips) -> Pips {
        match token_id.into() {
```

core/src/token_id/error.rs
```diff
@@ -0,0 +1,12 @@
+use super::MAX_ALLOWED_TOKEN_ID_LEN;
+use near_account_id::ParseAccountError;
+
+#[derive(thiserror::Error, Debug)]
+pub enum TokenIdError {
+    #[error("AccountId: {0}")]
+    AccountId(#[from] ParseAccountError),
+    #[error(transparent)]
+    ParseError(#[from] strum::ParseError),
+    #[error("TokenId is too long. Max length is {MAX_ALLOWED_TOKEN_ID_LEN}, got {0}")]
```
fusede, line 10: https://github.com/near/intents/pull/94#discussion_r2139574480

core/src/token_id/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    #[inline]
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (typ, data) = s
+            .split_once(':')
+            .ok_or(strum::ParseError::VariantNotFound)?;
+        Ok(match typ.parse()? {
+            TokenIdType::Nep141 => Self::Nep141(data.parse()?),
+            TokenIdType::Nep171 => Self::Nep171(data.parse()?),
+            TokenIdType::Nep245 => Self::Nep245(data.parse()?),
```
fusede, line 80:
```suggestion
            TokenIdType::Nep141 => data.parse().map(Self::Nep141),
            TokenIdType::Nep171 => data.parse().map(Self::Nep171),
            TokenIdType::Nep245 => data.parse().map(Self::Nep245),
```

core/src/token_id/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+)]
+#[strum_discriminants(
+    name(TokenIdType),
+    derive(strum::Display, EnumString),
+    strum(serialize_all = "snake_case"),
+    vis(pub)
+)]
+#[near(serializers = [borsh])]
+// Private: Because we need construction to go through the TokenId struct to check for length
+pub enum TokenId {
```
fusede, line 39: Can you please add test this enum de/serializes from/into borsh/json exactly like legacy enum that was before? You can define legacy enum in `mod tests`

core/src/token_id/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+    use arbitrary::{Arbitrary, Unstructured};
+    use near_sdk::{borsh, serde_json};
+    use rstest::rstest;
+    use test_utils::{
+        arbitrary::account_id::arbitrary_account_id,
+        asserts::ResultAssertsExt,
+        random::{Seed, gen_random_bytes, gen_random_string, make_seedable_rng, random_seed},
+    };
+
+    impl<'a> Arbitrary<'a> for TokenId {
```
fusede, line 146: Let's make it accessible not only for tests, so that this implementation can be re-used in other crates that depend on `defuse-core`.
Also, please gate it behind feature `arbitrary`:
```rust
#[cfg(feature = "arbitrary")]
const _: () = {
    impl<'a> Arbitrary<'a> for TokenId {
        // ...
    }
};
```

fusede (changes requested):

core/src/engine/state/cached.rs
BonelessImpl: Done

core/src/intents/token_diff.rs
BonelessImpl: If we do this, it can't be a const method.

core/src/token_id/mod.rs
BonelessImpl: In that case, how about I move it to test-utils? I do have a few other things there.

fusede: Rust doesn't allow you to implement traits on types that are defined in other crates

core/src/token_id/mod.rs
fusede: So, those hex values in the tests below were taken from serializing legacy `TokenId` enum? IMHO it would look more reliable if you copy-paste [legacy](https://github.com/near/intents/blob/852c34722e9dd26f563a374c9bc7dbd3028ecf47/core/src/tokens.rs#L35-L53) definition into `mod tests { ... }` and test with it explicitly (using `arbitrary` would be even better!), so we would eliminate human factor.

BonelessImpl: Done

PR #95, opened 2025-06-03
fix(ton_connect): check timestamp <= deadline

fusede:


core/src/payload/ton_connect.rs
```diff
@@ -25,6 +25,9 @@ where
     type Error = serde_json::Error;
 
     fn extract_defuse_payload(self) -> Result<DefusePayload<T>, Self::Error> {
+        // This was causing simulate_intents() to fail, since sometimes signed
+        // intent is simulated right after signing
+        #[cfg(feature = "ton_connect_check_timestamp")]
```
BonelessImpl, line 30: I'm going to approve this PR, but this is a very dangerous thing we're doing and I don't like it. That is: Using a feature for every little thing. Generally speaking, features hide code and creates multi-fold behavior that can't be easily reasoned about. Consider what happened with no_registration and how it multiplied the number of tests by two to ensure regression behavior.

To mathematically express this problem, we put it as follows. Using a feature is like setting a boolean over the whole code base. It's like making the whole source code a big struct with a state, and we have to wish that all future changes (and past changes) will not yield unexpected behavior to the whole source code when a certain value for the feature is chosen.

What I prefer is to include a localized boolean variable somewhere. Maybe in the payload. The boolean indicates whether checking the timestamp is necessary. At least that can be isolated and tested. We can think and come up with a better solution.

PR #103, opened 2025-06-10
feat: borsh-with

fusede:


borsh-utils/src/adapters/mod.rs
```diff
@@ -0,0 +1,463 @@
...
+    pub fn deserialize<R, U>(reader: &mut R) -> io::Result<U>
+    where
+        T: BorshDeserializeAs<U>,
+        R: io::Read,
+    {
+        T::deserialize_as(reader)
+    }
+}
+
+pub struct Same;
```
BonelessImpl, line 56: Can we call this "Identity"?

borsh-utils/src/adapters/mod.rs
```diff
@@ -0,0 +1,463 @@
...
+    mem::MaybeUninit,
+    rc::Rc,
+    sync::Arc,
+};
+
+use defuse_io_utils::ReadExt;
+use impl_tools::autoimpl;
+use near_sdk::borsh::{BorshDeserialize, BorshSerialize};
+
+#[cfg(feature = "chrono")]
```
BonelessImpl, line 16: Can we please remove these features? Having features in a sub crate in a workspace just makes it harder to integrate it with the IDE, leave alone that it becomes a burden to activate.

I know it's "nice"... looks fancy... but my god I hate these features! They just complicate builds and make it hard to ensure that all tests are running (given that I added tests in there).

borsh-utils/src/adapters/mod.rs
fusede: This is analog to [`serde_with::Same`](https://docs.rs/serde_with/latest/serde_with/struct.Same.html). I've added a comment about it

PR #106, opened 2025-06-19
feat: AML2

fusede:
## Summary by CodeRabbit

- **New Features**
  - Introduced comprehensive account locking and unlocking capabilities with role-based access control.
  - Enabled forced token transfers and withdrawals for privileged roles, bypassing account lock restrictions.
  - Added new event types to signal account lock and unlock actions.
  - Implemented versioned account serialization to support backward compatibility and future migrations.
- **Improvements**
  - Enhanced error variants to include detailed context for accounts and public keys.
  - Refined state management to enforce locking semantics with explicit error handling.
  - Streamlined intent execution and nonce management with improved error propagation.
  - Updated token transfer and withdrawal methods to support forced operations conditionally.
  - Extended test suites with integration tests covering account locking and legacy upgrades.
- **Bug Fixes**
  - Allowed deposits and incoming token transfers even when accounts are locked.
- **Documentation**
  - Added new traits and extended existing ones for forced operations and account locking, improving developer interfaces.
- **Chores**
  - Updated and reorganized dependencies and workspace configurations for development and testing efficiency.

### DIFF
--- a/defuse/src/contract/accounts/account/entry/mod.rs
+++ b/defuse/src/contract/accounts/account/entry/mod.rs
@@ -1,21 +1,21 @@
+mod v0;
+
 use std::{
     borrow::Cow,
     io::{self, Read},
     mem::size_of,
 };
 
-use defuse_bitmap::{U248, U256};
 use defuse_borsh_utils::adapters::{As, BorshDeserializeAs, BorshSerializeAs};
-use defuse_core::{Nonces, crypto::PublicKey};
+
 use defuse_near_utils::{Lock, PanicOnClone};
 use impl_tools::autoimpl;
 use near_sdk::{
     borsh::{BorshDeserialize, BorshSerialize},
     near,
-    store::{IterableSet, LookupMap},
 };
 
-use crate::contract::accounts::AccountState;
+use crate::contract::accounts::account::entry::v0::AccountV0;
 
 use super::Account;
 
@@ -138,41 +138,6 @@ impl BorshSerializeAs<Lock<Account>> for MaybeVersionedAccountEntry {
             VersionedAccountEntry::from(source),
         )
             .serialize(writer)
-    }
-}
-
-#[derive(Debug)]
-#[near(serializers = [borsh])]
-#[autoimpl(Deref using self.state)]
-#[autoimpl(DerefMut using self.state)]
-struct AccountV0 {
-    nonces: Nonces<LookupMap<U248, U256>>,
-
-    implicit_public_key_removed: bool,
-    public_keys: IterableSet<PublicKey>,
-
-    pub state: AccountState,
-
-    prefix: Vec<u8>,
-}
-
-impl From<AccountV0> for Account {
-    fn from(
-        AccountV0 {
-            nonces,
-            implicit_public_key_removed,
-            public_keys,
-            state,
-            prefix,
-        }: AccountV0,
-    ) -> Self {
-        Self {
-            nonces,
-            implicit_public_key_removed,
-            public_keys,
-            state,
-            prefix,
-        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


fusede:
@coderabbitai review

fusede:
@coderabbitai review

bitmap/src/lib.rs
```diff
@@ -76,6 +77,28 @@ where
             self.clear_bit(n)
         }
     }
+
+    /// Iterate over set U256
+    #[inline]
+    pub fn iter(&self) -> impl Iterator<Item = U256> + '_
```
BonelessImpl, line 83: Perhaps this should be called "as_iter"? It's unlike both into_iter and iter. It takes a reference and returns values.

near-utils/src/lock.rs
```diff
@@ -35,81 +41,203 @@ impl<T> Lock<T> {
...
+
+    #[must_use]
+    #[inline]
+    pub const fn as_locked_or_mut(&mut self, force: bool) -> Option<&mut T> {
+        if force {
+            Some(self.as_inner_unchecked_mut())
+        } else {
+            self.as_locked_mut()
+        }
     }
```
BonelessImpl, line 104: I bet these functions will never be used. The default usage is always when something is unlocked. When we want to access something and assert it's locked, we can do it on-spot. How many times were any of these used? I bet maybe once at most. And I haven't checked before writing the comment.

Instead of this major complexity in naming and usage. Why don't we simply remove these functions, and change "as_unlocked" to "get_*", which is Rust's idiomatic way of naming things? Anyone reading "get" and seeing it returns Option will immediately understand what this struct is doing. But having 10 different variants of "as"... too much! We don't need all this.

I really have to admit that the terminology and way this is done is way too complex and hard to understand.

tests/src/tests/defuse/accounts/mod.rs
```diff
@@ -42,6 +53,24 @@ impl AccountManagerExt for near_workspaces::Account {
         Ok(())
     }
 
+    async fn remove_public_key(
+        &self,
+        defuse_contract_id: &AccountId,
+        public_key: PublicKey,
+    ) -> anyhow::Result<()> {
+        // TODO: check bool output
```
BonelessImpl, line 61: There are no more bool outputs, right? it's now a Result.

tests/src/tests/defuse/accounts/locked.rs
```diff
@@ -0,0 +1,446 @@
...
+            }))
+            .deposit(NearToken::from_yoctonear(1))
+            .max_gas()
+            .transact()
+            .await?
+            .into_result()?
+            .json()
+            .map_err(Into::into)
+    }
+}
```
BonelessImpl, line 446: It would be nice if traits and impls can be moved to a separate module, where only tests are separate from impl details.

near-utils/src/lock.rs
```diff
@@ -35,81 +41,203 @@ impl<T> Lock<T> {
...
+        if self.is_locked() {
             return None;
         }
-        Some(&self.value)
+        Some(self.as_inner_unchecked_mut())
     }
 
+    #[must_use]
+    #[inline]
+    pub const fn as_unlocked_or_mut(&mut self, force: bool) -> Option<&mut T> {
```
BonelessImpl, line 151: Following up on my other comment in this file, this would become `as_mut`... much cleaner!

defuse/src/contract/accounts/lock.rs
```diff
@@ -0,0 +1,46 @@
...
+#[near]
+impl AccountForceLocker for Contract {
+    fn is_account_locked(&self, account_id: &AccountId) -> bool {
+        StateView::is_account_locked(self, account_id)
+    }
+
+    #[access_control_any(roles(Role::DAO, Role::UnrestrictedAccountLocker))]
+    #[payable]
+    fn force_lock_account(&mut self, account_id: AccountId) -> bool {
+        assert_one_yocto();
```
BonelessImpl, line 20: Do you think it's worth it to write some software that will automatically check for invariants, like one_yocto, and fail if these invariants are not satisfied? I'm always concerned about APIs that have such requirements and are manually checked.

BonelessImpl (commented):
First round done.

PR #131, opened 2025-09-30
Salted nonces

hlgltvnnk:
- Nonce versioning
- Salted nonces
- State versioning


core/src/error.rs
```diff
@@ -66,4 +69,7 @@ pub enum DefuseError {
 
     #[error("wrong verifying_contract")]
     WrongVerifyingContract,
+
+    #[error("invalid salt")]
+    InvalidSalt,
```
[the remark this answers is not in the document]
hlgltvnnk: InvalidSalt appears not only when checking the nonce, but also when invalidating the salt (if such a salt does not exist - error), so it should not be reduced only to the nonce.

core/src/engine/mod.rs
```diff
@@ -87,6 +84,34 @@ where
         Ok(())
     }
 
+    #[inline]
+    fn verify_intent_nonce(&self, nonce: Nonce, intent_deadline: Deadline) -> Result<()> {
+        let versioned = VersionedNonce::try_from(nonce).map_err(|_| DefuseError::InvalidNonce)?;
```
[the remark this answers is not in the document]
hlgltvnnk: upd: changed to From<...> - added panic to implementation, any errors should occur in tests

defuse/src/contract/accounts/mod.rs
```diff
@@ -52,15 +52,23 @@ impl AccountManager for Contract {
...
-            )
-            .ok();
+            for nonce in nonces.into_iter().map(AsBase64::into_inner) {
+                let cleanable = match VersionedNonce::from(nonce) {
+                    VersionedNonce::V1(SaltedNonce {
+                        salt,
+                        nonce: ExpirableNonce { deadline, .. },
+                    }) => !self.is_valid_salt(salt) || deadline.has_expired(),
+                    // NOTE: legacy nonces can't be cleared before a complete prohibition on its usage
+                    VersionedNonce::Legacy(_) => false,
```
fusede, line 64:
```suggestion
                    VersionedNonce::Legacy(_) => continue,
```

core/src/nonce/salted.rs
```diff
@@ -0,0 +1,175 @@
...
+        Self(result)
+    }
+}
+
+/// Contains current valid salt and set of previous
+/// salts that can be valid or invalid.
+#[near(serializers = [borsh])]
+#[derive(Debug)]
+pub struct ValidSalts {
+    previous: IterableMap<Salt, bool>,
```
fusede, line 44:
```suggestion
    previous: IterableMap<Salt, bool, Identity>,
```

core/src/nonce/versioned.rs
```diff
@@ -0,0 +1,133 @@
...
+pub enum VersionedNonce {
+    Legacy(Nonce),
+    V1(SaltedNonce<ExpirableNonce<[u8; 15]>>),
+}
+
+// Allowed to maintain a clean API
+#[allow(clippy::fallible_impl_from)]
+impl From<Nonce> for VersionedNonce {
+    fn from(value: Nonce) -> Self {
+        MaybeVersionedNonce::deserialize_as(&mut value.as_ref()).unwrap()
```
fusede, line 32:
```suggestion
        MaybeVersionedNonce::deserialize_as(&mut value.as_ref()).unwrap_or_else(|_| unreachable!())
```

defuse/src/contract/accounts/account/mod.rs
```diff
@@ -149,12 +149,12 @@ impl Account {
...
 
-    /// Clears the nonce if it was expired.
-    /// Returns whether the nonces was cleared. If the nonce has not expired yet, then returns `false`,
+    /// Clears the nonce if it was expired/invalidated.
+    /// Returns whether the nonces was cleared,
     /// regardless of whether it was previously committed or not.
     #[inline]
-    pub fn clear_expired_nonce(&mut self, nonce: U256) -> bool {
-        self.nonces.clear_expired(nonce)
+    pub fn cleanup_nonce(&mut self, nonce: U256) -> bool {
```
fusede, line 156:
```suggestion
    pub fn cleanup_nonce_by_prefix(&mut self, prefix: [u8; 31]) -> bool {
```

defuse/src/contract/accounts/mod.rs
```diff
@@ -52,15 +52,23 @@ impl AccountManager for Contract {
...
-                &account_id,
-                nonces.into_iter().map(AsBase64::into_inner),
-            )
-            .ok();
+            for nonce in nonces.into_iter().map(AsBase64::into_inner) {
+                let cleanable = match VersionedNonce::from(nonce) {
+                    VersionedNonce::V1(SaltedNonce {
+                        salt,
+                        nonce: ExpirableNonce { deadline, .. },
+                    }) => !self.is_valid_salt(salt) || deadline.has_expired(),
```
fusede, line 62:
```suggestion
                    }) => deadline.has_expired() || !self.is_valid_salt(salt),
```

core/src/engine/state/cached.rs
```diff
@@ -417,7 +415,7 @@ impl CachedAccount {
     }
 
     #[inline]
-    pub fn clear_expired_nonce(&mut self, n: U256) -> bool {
-        self.nonces.clear_expired(n)
+    pub fn cleanup_nonce_by_prefix(&mut self, nonce: Nonce) -> bool {
```
fusede, line 418: Let's be explicit about what prefix we're cleaning up by:
```suggestion
    pub fn cleanup_nonce_by_prefix(&mut self, prefix: [u8; 31]) -> bool {
```
So that the caller cannot make wrong assumptions on the length of the prefix that will be used

core/src/nonce/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub fn commit(&mut self, n: Nonce) -> Result<()> {
+        if self.0.set_bit(n) {
+            return Err(DefuseError::NonceUsed);
+        }
+
+        Ok(())
+    }
+
+    #[inline]
+    pub fn cleanup_by_prefix(&mut self, nonce: Nonce) -> bool {
```
fusede, line 50: And most importantly, here
```suggestion
    pub fn cleanup_by_prefix(&mut self, prefix: [u8; 31]) -> bool {
```

core/src/nonce/versioned.rs
```diff
@@ -0,0 +1,122 @@
...
+        let mut prefix = [0u8; 4];
+        reader.read_exact(&mut prefix)?;
+
+        let versioned = if prefix == Self::VERSIONED_MAGIC_PREFIX {
+            VersionedNonce::deserialize_reader(reader)?
+        } else {
+            VersionedNonce::Legacy(Nonce::deserialize_reader(&mut prefix.chain(reader))?)
+        };
+
+        Ok(versioned)
```
fusede, line 68: Nit: a bit cleaner to not use `?` and then wrap into `Ok` again IMO:
```suggestion
        if prefix == Self::VERSIONED_MAGIC_PREFIX {
            VersionedNonce::deserialize_reader(reader)
        } else {
            Nonce::deserialize_reader(&mut prefix.chain(reader)).map(VersionedNonce::Legacy)
        }
```

defuse/src/contract/accounts/account/mod.rs
```diff
@@ -149,12 +149,12 @@ impl Account {
...
 
-    /// Clears the nonce if it was expired.
-    /// Returns whether the nonces was cleared. If the nonce has not expired yet, then returns `false`,
+    /// Clears the all nonces with corresponding prefix if it was expired/invalidated.
+    /// Returns whether the nonces was cleared,
     /// regardless of whether it was previously committed or not.
     #[inline]
-    pub fn clear_expired_nonce(&mut self, nonce: U256) -> bool {
-        self.nonces.clear_expired(nonce)
+    pub fn cleanup_nonce_by_prefix(&mut self, nonce: Nonce) -> bool {
```
fusede, line 156: Be explicit about pefix:
```suggestion
    pub fn cleanup_nonce_by_prefix(&mut self, prefix: [u8; 31]) -> bool {
```

defuse/src/contract/salts.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        let current_salt = self.salts.current();
+
+        SaltRotationEvent {
+            current: current_salt,
+            invalidated,
+        }
+        .emit();
+
+        current_salt
```
fusede, line 54: Let's skip errors on already invalidated or non-existing salts. It's also more preferrable to avoid `mut` where possible and collect from iterators instead (less chance of making a mistake):
```suggestion
        let invalidated = salts
            .into_iter()
            .filter(|s| self.salts.invalidate(s))
            .collect();
        let current = self.salts.current();

        SaltRotationEvent {
            current,
            invalidated,
        }
        .emit();

        current
```

defuse/src/contract/state/v0.rs
```diff
@@ -0,0 +1,64 @@
...
+        prefix: S,
+    ) -> Self
+    where
+        S: IntoStorageKey,
+    {
+        Self {
+            total_supplies,
+            wnear_id,
+            fees,
+            salts: SaltRegistry::new(prefix.into_storage_key().nest(Prefix::Salts)),
```
fusede, line 36: Are you sure about `.into_storage_key()`?

core/src/nonce/versioned.rs
```diff
@@ -0,0 +1,122 @@
...
+        let mut result = [0u8; SIZE];
+
+        MaybeVersionedNonce::serialize_as(&value, &mut result.as_mut_slice())
+            .unwrap_or_else(|_| unreachable!());
+
+        result
+    }
+}
+
+struct MaybeVersionedNonce;
```
fusede, line 47: I've just noticed that it might be cleaner if we move `Legacy` variant from `VersionedNonce` into `MaybeVersionedNonce`:
```rust
pub enum MaybeVersionedNonce {
    Legacy(Nonce),
    Versioned(VersionedNonce)
}

impl BorshSerialize for MaybeVersionedNonce { /* ... */ }
impl BorshDeserialize for MaybeVersionedNonce { /* ... */ }
impl From<Nonce> for MaybeVersionedNonce { /* ... */ }
impl From<MaybeVersionedNonce> for Nonce { /* ... */ }
```

It will help us to get rid of `Legacy` variant afterwards (when we rotate salt) and just start using `VersionedNonce` alone: `Legacy` variant would be just not there.

PR #138, opened 2025-10-20
Test/migration testing

hlgltvnnk:
Changed:
Separated token and user creation from environment setup

Added:
State migration testing:
- via a separate mod in the environment variable
- dedicated tests for data consistency

Environment setup flow for migration tests:
1. Deploy the old version of the contract
2. Generate test data (users, tokens, balances, etc.)
3. Apply test data
4. Migrate the contract to the new version
5. Check the consistency of old data

Creating users in migration mode: users are created/extracted from previously created ones to verify the interaction of old and new data during migration

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed `cleanup_nonces` method from the `AccountManager` trait implementation for `Contract`, eliminating per-account nonce cleanup functionality

### DIFF
--- a/tests/src/tests/defuse/mod.rs
+++ b/tests/src/tests/defuse/mod.rs
@@ -1,5 +1,6 @@
 pub mod accounts;
 mod env;
+mod garbage_collector;
 mod intents;
 mod state;
 mod storage;
--- a/defuse/src/accounts.rs
+++ b/defuse/src/accounts.rs
@@ -29,11 +29,6 @@ pub trait AccountManager {
     /// [permit2 nonce schema](https://docs.uniswap.org/contracts/permit2/reference/signature-transfer#nonce-schema).
     fn is_nonce_used(&self, account_id: &AccountId, nonce: AsBase64<Nonce>) -> bool;
 
-    /// Clears all expired nonces for given accounts by its prefix.
-    /// Omitting any errors, e.g. if account doesn't exist or nonces are not expired.
-    /// NOTE: MUST attach 1 yⓃ for security purposes.
-    fn cleanup_nonces(&mut self, nonces: Vec<(AccountId, Vec<AsBase64<Nonce>>)>);
-
     /// Returns whether authentication by PREDECESSOR_ID is enabled
     /// for given `account_id`.
     ///
--- a/defuse/src/contract/accounts/mod.rs
+++ b/defuse/src/contract/accounts/mod.rs
@@ -7,7 +7,7 @@ pub use self::{account::*, state::*};
 use std::{borrow::Cow, collections::HashSet};
 
 use defuse_core::{
-    DefuseError, ExpirableNonce, Nonce, SaltedNonce, VersionedNonce,
+    DefuseError, Nonce,
     accounts::{AccountEvent, PublicKeyEvent},
     crypto::PublicKey,
     engine::{State, StateView},
@@ -17,7 +17,6 @@ use defuse_core::{
 use defuse_near_utils::{Lock, NestPrefix, PREDECESSOR_ACCOUNT_ID, UnwrapOrPanic};
 use defuse_serde_utils::base64::AsBase64;
 
-use near_plugins::{AccessControllable, access_control_any};
 use near_sdk::{
     AccountId, AccountIdRef, BorshStorageKey, FunctionError, IntoStorageKey, assert_one_yocto,
     borsh::BorshSerialize, near, store::IterableMap,
@@ -25,7 +24,7 @@ use near_sdk::{
 
 use crate::{
     accounts::AccountManager,
-    contract::{Contract, ContractExt, Role, accounts::AccountEntry},
+    contract::{Contract, ContractExt, accounts::AccountEntry},
 };
 
 #[near]
@@ -72,24 +71,6 @@ impl AccountManager for Contract {
         StateView::is_nonce_used(self, account_id, nonce.into_inner())
     }
 
-    #[access_control_any(roles(Role::DAO, Role::GarbageCollector))]
-    #[payable]
-    fn cleanup_nonces(&mut self, nonces: Vec<(AccountId, Vec<AsBase64<Nonce>>)>) {
-        assert_one_yocto();
-
-        for (account_id, nonces) in nonces {
-            for nonce in nonces.into_iter().map(AsBase64::into_inner) {
-                if !self.is_nonce_cleanable(nonce) {
-                    continue;
-                }
-
-                // NOTE: all errors are omitted
-                let [prefix @ .., _] = nonce;
-                let _ = State::cleanup_nonce_by_prefix(self, &account_id, prefix);
-            }
-        }
-    }
-
     fn is_auth_by_predecessor_id_enabled(&self, account_id: &AccountId) -> bool {
         StateView::is_auth_by_predecessor_id_enabled(self, account_id)
     }
@@ -103,20 +84,6 @@ impl AccountManager for Contract {
 }
 
 impl Contract {
-    #[inline]
-    fn is_nonce_cleanable(&self, nonce: Nonce) -> bool {
-        let Some(versioned_nonce) = VersionedNonce::maybe_from(nonce) else {
-            return false;
-        };
-
-        match versioned_nonce {
-            VersionedNonce::V1(SaltedNonce {
-                salt,
-                nonce: ExpirableNonce { deadline, .. },
-            }) => deadline.has_expired() || !self.is_valid_salt(salt),
-        }
-    }
-
     #[inline]
     pub fn ensure_auth_predecessor_id(&self) -> &'static AccountId {
         if !StateView::is_auth_by_predecessor_id_enabled(self, &PREDECESSOR_ACCOUNT_ID) {
--- a/defuse/src/contract/garbage_collector.rs
+++ b/defuse/src/contract/garbage_collector.rs
@@ -1,0 +1,44 @@
+use defuse_core::{ExpirableNonce, Nonce, SaltedNonce, VersionedNonce, engine::State};
+use defuse_serde_utils::base64::AsBase64;
+use near_plugins::{AccessControllable, access_control_any};
+use near_sdk::{AccountId, assert_one_yocto, near};
+
+use super::{Contract, ContractExt, Role};
+use crate::{garbage_collector::GarbageCollector, salts::SaltManager};
+
+#[near]
+impl GarbageCollector for Contract {
+    #[access_control_any(roles(Role::DAO, Role::GarbageCollector))]
+    #[payable]
+    fn cleanup_nonces(&mut self, nonces: Vec<(AccountId, Vec<AsBase64<Nonce>>)>) {
+        assert_one_yocto();
+
+        for (account_id, nonces) in nonces {
+            for nonce in nonces.into_iter().map(AsBase64::into_inner) {
+                if !self.is_nonce_cleanable(nonce) {
+                    continue;
+                }
+
+                // NOTE: all errors are omitted
+                let [prefix @ .., _] = nonce;
+                let _ = State::cleanup_nonce_by_prefix(self, &account_id, prefix);
+            }
+        }
+    }
+}
+
+impl Contract {
+    #[inline]
+    fn is_nonce_cleanable(&self, nonce: Nonce) -> bool {
+        let Some(versioned_nonce) = VersionedNonce::maybe_from(nonce) else {
+            return false;
+        };
+
+        match versioned_nonce {
+            VersionedNonce::V1(SaltedNonce {
+                salt,
+                nonce: ExpirableNonce { deadline, .. },
+            }) => deadline.has_expired() || !self.is_valid_salt(salt),
+        }
+    }
+}
--- a/defuse/src/contract/mod.rs
+++ b/defuse/src/contract/mod.rs
@@ -5,6 +5,7 @@ mod admin;
 pub mod config;
 mod events;
 mod fees;
+mod garbage_collector;
 mod intents;
 mod salts;
 mod state;
--- a/defuse/src/garbage_collector.rs
+++ b/defuse/src/garbage_collector.rs
@@ -1,0 +1,12 @@
+use defuse_core::Nonce;
+use defuse_serde_utils::base64::AsBase64;
+use near_sdk::{AccountId, ext_contract};
+
+#[ext_contract(ext_garbage_collector)]
+#[allow(clippy::module_name_repetitions)]
+pub trait GarbageCollector {
+    /// Clears all expired nonces for given accounts by its prefix.
+    /// Omitting any errors, e.g. if account doesn't exist or nonces are not expired.
+    /// NOTE: MUST attach 1 yⓃ for security purposes.
+    fn cleanup_nonces(&mut self, nonces: Vec<(AccountId, Vec<AsBase64<Nonce>>)>);
+}
--- a/defuse/src/lib.rs
+++ b/defuse/src/lib.rs
@@ -4,6 +4,7 @@ pub mod contract;
 
 pub mod accounts;
 pub mod fees;
+pub mod garbage_collector;
 pub mod intents;
 pub mod salts;
 pub mod simulation_output;
--- a/io-utils/src/lib.rs
+++ b/io-utils/src/lib.rs
@@ -42,8 +42,9 @@ where
 
     /// Reads all remaining data into a buffer and writes it to the internal writer.
     fn read_to_end(&mut self, buf: &mut Vec<u8>) -> Result<usize> {
+        let len = buf.len();
         let n = self.reader.read_to_end(buf)?;
-        self.writer.write_all(&buf[..n])?;
+        self.writer.write_all(&buf[len..])?;
         Ok(n)
     }
 }
--- a/tests/src/tests/defuse/accounts/mod.rs
+++ b/tests/src/tests/defuse/accounts/mod.rs
@@ -9,8 +9,6 @@ use defuse_serde_utils::base64::AsBase64;
 use near_sdk::{AccountId, AccountIdRef, Gas, NearToken};
 use serde_json::json;
 
-use crate::utils::test_log::TestLog;
-
 pub trait AccountManagerExt {
     async fn add_public_key(
         &self,
@@ -23,12 +21,6 @@ pub trait AccountManagerExt {
         defuse_contract_id: &AccountId,
         public_key: PublicKey,
     ) -> anyhow::Result<()>;
-
-    async fn cleanup_nonces(
-        &self,
-        defuse_contract_id: &AccountId,
-        data: &[(AccountId, Vec<Nonce>)],
-    ) -> anyhow::Result<TestLog>;
 
     async fn defuse_has_public_key(
         &self,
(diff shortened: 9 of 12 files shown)


tests/src/tests/defuse/accounts/auth_by_predecessor_id.rs
```diff
@@ -19,19 +20,29 @@ use crate::{
...
 #[tokio::test]
 #[rstest]
-async fn test(random_bytes: Vec<u8>) {
+async fn auth_by_predecessor_id(random_bytes: Vec<u8>) {
     let mut u = Unstructured::new(&random_bytes);
-    let env = Env::new().await;
+    let mut env = Env::new().await;
+
+    let ft = env.create_token().await;
+    let user = env.get_or_create_user().await;
```
fusede: Test implementations shouldn't know whether these users were created before (for migration testing) or not, so maybe better to avoid confusion here and rename to `Env::create_user()`?
```suggestion
    let user = env.create_user().await;
```

tests/src/tests/defuse/mod.rs
```diff
@@ -22,19 +23,40 @@ use defuse::{
...
 };
 use near_sdk::{AccountId, serde::Serialize, serde_json::json};
+use near_sdk::{Gas, NearToken};
 use near_workspaces::Contract;
 use std::sync::LazyLock;
 
-static DEFUSE_WASM: LazyLock<Vec<u8>> = LazyLock::new(|| read_wasm("defuse"));
+static DEFUSE_WASM: LazyLock<Vec<u8>> = LazyLock::new(|| read_wasm("res", "defuse"));
+static DEFUSE_LEGACY_WASM: LazyLock<Vec<u8>> =
+    LazyLock::new(|| read_wasm("releases", "0.2.10.wasm"));
```
fusede: hmm, maybe then rename file so that it points to the "previous version", not a specific one?

tests/src/tests/defuse/env/builder.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        env
+    }
+
+    pub async fn build_with_migration(&mut self) -> Env {
+        self.build_env(true).await
+    }
+
+    pub async fn build(&mut self) -> Env {
+        let migrate_from_legacy = std::env::var("MIGRATE_FROM_LEGACY").is_ok_and(|v| v != "0");
```
fusede:
```suggestion
        let migrate_from_legacy = std::env::var("MIGRATE_FROM_LEGACY").is_ok_and(|v| !["0", "false"].contains(v.to_lower()));
```

tests/src/tests/defuse/env/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+        ExpirableNonce {
+            deadline,
+            nonce: rng.random::<[u8; 15]>(),
+        },
+    ))
+    .into()
+}
+
+fn generate_random_account_id(parent_id: &AccountId) -> AccountId {
+    let mut rng = make_true_rng();
```
fusede: The idea behind using `make_true_rng` once is that it prints current seed value that was used for the whole test, so that these tests are reproducible: if we ever catch an error with some specific seed value, then we can re-run the same test by manually specifying the seed, so that we will have the same result. Do you think we should deviate from this approach here or can you suggest an alternative?

tests/src/tests/defuse/env/storage.rs
```diff
@@ -0,0 +1,217 @@
...
+                    .unwrap();
+
+                assert!(has_key);
+            }
+
+            for nonce in &data.nonces {
+                let is_used = self.defuse.is_nonce_used(account_id, nonce).await.unwrap();
+                assert!(is_used);
+            }
+        }
```
fusede: everything here can be done concurrently

tests/src/tests/defuse/env/storage.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    async fn apply_account(&self, data: (&AccountId, &AccountData)) -> Result<Account> {
+        let (account_id, account) = data;
+        let acc = self
+            .create_named_user(&self.sandbox.subaccount_name(account_id))
+            .await?;
+
+        self.apply_public_keys(&acc, account).await?;
+
+        self.apply_nonces(&acc, account).await?;
```
fusede: this can be done concurrently

fusede: PS: adding public keys is implemented via intents, so it also commits some nonces. Do we still need to commit a separate set of nonces?

tests/src/tests/defuse/env/storage.rs
```diff
@@ -0,0 +1,217 @@
...
+        },
+        poa::factory::PoAFactoryExt,
+    },
+    utils::{ParentAccount, ft::FtExt, mt::MtExt},
+};
+
+pub trait StorageMigration {
+    async fn generate_storage_data(&mut self);
+    async fn verify_storage_consistency(&self);
+}
```
fusede: do you really need this to be a trait?

tests/src/tests/defuse/env/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+                .await
+                .unwrap();
+
+            // FT storage deposit
+            for token in tokens {
+                self.poa_factory
+                    .ft_storage_deposit_many(token, &accounts)
+                    .await
+                    .unwrap();
+            }
```
fusede: this can be done concurrently

tests/src/tests/defuse/env/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+                self.poa_factory.id(),
+                &self.poa_ft_name(ft),
+                root.id(),
+                1_000_000_000,
+                None,
+                None,
+            )
+            .await
+            .unwrap();
+        }
```
fusede: this can be done concurrently

fusede (changes requested):
first iteration

tests/src/tests/defuse/mod.rs
hlgltvnnk: it is better to leave a dedicated version so that we know exactly which version contained the previous state version of the contract, and it will also reduce the number of human errors when replacing wasms during new migrations

pityjllk (commented):
I really like the test framework changes like on demand account/tokens creatikon. Also i have ew thoughts i had while reviewing the pr

- i personally think that dedicated test cases for migration would suffice. Running all existing test cases on upgraded state can still miss some corner cases so since we need to have dedicated test cases for storage migrations ... and since we need them then maybe running test cases on upgraded state doesnt bring that match to the table... (the argument i made up is really weak, but just to give an idea how it feels like for me 🙃 ) 

- im more in favour of having single test cases covering particular scenario rather some complicated setup upfront. Similiarly you could maybe argure that for each account that we want to use for smart contract call we could have a flag that will enable some extra action to be executed on that account so it s more like blockchain account rather than fresh one... But at the end of the day you still need to test both scenarios (fresh and old account ) .... 

- i am not sure how will we approach future migrations... One could argue that we should only test latest migration (current => next) but theoretically there could be scenarios that wont be cover by migration `B => C` while theoretically could be covered by `A => B => C`. So if we can skip `A => B` i would argue that we can skip `B => C` and just have dedicated tests for migration.

- i personally think that we should cover data migration in separate test case. And since all the effort is focused on making sure that migration will suceed on contract state from mainnet i would give a try for https://github.com/near/near-workspaces-rs/blob/main/workspaces/tests/patch_state.rs#L104-L108

tests/src/tests/defuse/env/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+
+    pub poa_factory: Contract,
+
+    pub disable_ft_storage_deposit: bool,
+    pub disable_registration: bool,
+
+    // Persistent state generated in case of migration tests
+    // used to fetch existing accounts
+    pub persistent_state: Option<PersistentState>,
+    pub current_user_index: usize,
```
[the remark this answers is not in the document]
hlgltvnnk: Used mutex instead of AtomicUsize to ensure both checking and increment happen atomically. With atomic usize, there's a race condition between checking the length and incrementing the counter  which can be solved with compare_exchange loop, so i decided to use mutex for simpler code with negligible performance overhead for this use case

tests/src/tests/defuse/env/builder.rs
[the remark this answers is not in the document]
hlgltvnnk: build_with_migration was added only for 1 dedicated test which covers migrated state verification regardless of whether MIGRATE_FROM_LEGACY enabled or not. 
So each time ci runs tests state migration is verified by test_upgrade_with_persistence, but when MIGRATE_FROM_LEGACY var enabled migration is applied to all tests setup  allowing reuse of already created data, combining old and new data in the test, thus running migration tests on the entire functionality, not just on its individual parts
We can enable MIGRATE_FROM_LEGACY on ci and run state migration tests each time - and on the one hand, this may be unnecessary, since migration tests only make sense when the contract state version changes, but on the other hand, it would reduce human errors during development

tests/src/tests/defuse/env/builder.rs
```diff
@@ -0,0 +1,184 @@
...
+
+        env
+    }
+
+    pub async fn build_with_migration(&mut self) -> Env {
+        self.build_env(true).await
+    }
+
+    pub async fn build(&mut self) -> Env {
+        let migrate_from_legacy = std::env::var("MIGRATE_FROM_LEGACY")
```
mitinarseny:
```rust
const MIGRATE_FROM_LEGACY_ENV_NAME: &str = "DEFUSE_MIGRATE_FROM_LEGACY";
```

Makefile.toml
```diff
@@ -94,16 +94,34 @@ args = [
 ]
 
 [tasks.test]
-alias = "tests"
+dependencies = ["build"]
+run_task = "tests"
```
fusede: what's the difference?

tests/src/tests/defuse/state/fee.rs
```diff
@@ -12,12 +12,12 @@ use crate::{
...
 async fn set_fee() {
-    let mut env = Env::builder().deployer_as_super_admin().build().await;
+    let env = Env::builder().deployer_as_super_admin().build().await;
     let prev_fee = env.defuse.fee(env.defuse.id()).await.unwrap();
     let fee = Pips::from_pips(100).unwrap();
 
-    let user1 = env.get_or_create_user().await;
-    let user2 = env.get_or_create_user().await;
+    let user1 = env.create_user().await;
+    let user2 = env.create_user().await;
```
fusede: this can be done concurrently

tests/src/tests/defuse/state/salt.rs
```diff
@@ -11,11 +11,11 @@ use crate::{
...
 #[rstest]
 async fn update_current_salt() {
-    let mut env = Env::builder().deployer_as_super_admin().build().await;
+    let env = Env::builder().deployer_as_super_admin().build().await;
     let prev_salt = env.defuse.current_salt(env.defuse.id()).await.unwrap();
 
-    let user1 = env.get_or_create_user().await;
-    let user2 = env.get_or_create_user().await;
+    let user1 = env.create_user().await;
+    let user2 = env.create_user().await;
```
fusede: this can be done concurrently

tests/src/tests/defuse/storage/mod.rs
```diff
@@ -34,14 +34,14 @@ async fn storage_deposit_success(
...
+    let env = Env::builder()
         .disable_ft_storage_deposit()
         .no_registration(false)
         .build()
         .await;
 
-    let user = env.get_or_create_user().await;
-    let other_user = env.get_or_create_user().await;
+    let user = env.create_user().await;
+    let other_user = env.create_user().await;
```
fusede: this can be done concurrently

tests/src/tests/defuse/intents/simulate.rs
```diff
@@ -545,30 +574,40 @@ async fn simulate_token_diff_intent(
...
-            .mt_balance_of(env.user1.id(), &ft1_token_id.to_string())
+            .mt_balance_of(user1.id(), &ft1_token_id.to_string())
             .await
             .unwrap(),
         100
     );
     assert_eq!(
         env.defuse
-            .mt_balance_of(env.user2.id(), &ft2_token_id.to_string())
+            .mt_balance_of(user2.id(), &ft2_token_id.to_string())
```
fusede: this can be done concurrently

tests/src/tests/defuse/intents/simulate.rs
```diff
@@ -877,9 +924,8 @@ async fn simulate_auth_call_intent(nonce: Nonce, signing_standard: SigningStanda
...
         result.report.logs,
-        vec![
-            AccountNonceIntentEvent::new(&env.user1.id(), nonce, &auth_call_payload)
-                .into_event_log(),
-        ]
+        vec![AccountNonceIntentEvent::new(&user1.id(), nonce, &auth_call_payload).into_event_log(),]
     );
 }
+
+// ::intents::simulate::simulate_mt_withdraw_intent ... FAILED
```
fusede:
```suggestion
```

tests/src/tests/defuse/env/storage.rs
```diff
@@ -171,47 +153,94 @@ impl Env {
...
+
+        Ok(())
+    }
+
+    async fn check_balance(
+        &self,
+        account_id: &AccountId,
+        token_id: &Nep141TokenId,
+        amount: u128,
+    ) -> Result<()> {
```
fusede:
```suggestion
    async fn check_nep141_balance_eq(
        &self,
        account_id: &AccountId,
        token_id: Nep141TokenId,
        amount: u128,
    ) -> Result<()> {
```

tests/src/tests/defuse/env/storage.rs
```diff
@@ -171,47 +153,94 @@ impl Env {
...
+                .iter()
+                .map(|(token_id, expected_balances)| {
+                    self.verify_token_balance(token_id, expected_balances)
+                }),
+        )
+        .await
+        .expect("Failed to verify token balances");
+    }
+
+    async fn verify_token_balance(
```
fusede:
```suggestion
    async fn check_nep141_balances_eq(
```

tests/src/tests/defuse/env/storage.rs
```diff
@@ -25,54 +25,64 @@ use crate::{
...
+
+        try_join_all(state.accounts.iter().map(|data| self.apply_account(data)))
+            .await
+            .expect("Failed to apply accounts");
+    }
+
+    async fn apply_token_balances(&self) {
+        let state = self.state();
+
+        // NOTE: should be sequential to match contract token ordering
```
fusede: what do you mean by "contract token ordering"?

tests/src/tests/defuse/env/mod.rs
```diff
@@ -65,6 +64,12 @@ impl Env {
         Self::builder().build().await
     }
 
+    pub fn state(&self) -> &PersistentState {
+        self.persistent_state
+            .as_ref()
+            .expect("Persistent state is not set")
+    }
+
```
fusede: The only use of it externally is in `test_upgrade_with_persistence` test, check out the comment below how to avoid having this method in the first place.

fusede (changes requested):

Makefile.toml
```diff
@@ -94,16 +94,34 @@ args = [
...
+    1|true|TRUE|yes|YES)
+        echo "Migration Mode: ENABLED"
+        ;;
+    *)
+        echo "Migration Mode: DISABLED"
+        ;;
+esac
+
+echo "--------------------------------------------"
+'''
```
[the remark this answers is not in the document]
hlgltvnnk: then it will be duplicated every time in each test, but added information to the makefile  is printed only once

tests/src/tests/defuse/env/mod.rs
```diff
@@ -144,28 +149,24 @@ impl Env {
...
+    async fn get_next_account_id(&self) -> AccountId {
         if let Some(state) = self.persistent_state.as_ref() {
-            if self.current_user_index < state.accounts.len() {
-                let account_id = state
-                    .accounts
-                    .keys()
-                    .nth(self.current_user_index)
-                    .expect("Failed to get account ID")
-                    .clone();
+            let mut index = self.current_user_index.lock().await;
```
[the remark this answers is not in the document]
pityjllk: maybe this https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicI64.html#method.fetch_add ?

i think at this point state is already pregenrated(?) and static, so its enought to serve every individual caller unique index 🤔  (so some will get pregenerated accounts and others will generate new onces)

tests/src/tests/defuse/intents/simulate.rs
```diff
@@ -874,21 +874,18 @@ async fn simulate_auth_call_intent(nonce: Nonce, signing_standard: SigningStanda
...
-    user1
-        .near_deposit(env.wnear.id(), wnear_amount)
-        .await
-        .unwrap();
 
-    user1
-        .ft_transfer_call(
+    futures::try_join!(
+        user1.near_deposit(env.wnear.id(), wnear_amount),
+        user1.ft_transfer_call(
```
fusede: this should be sequential, I guess

tests/src/tests/defuse/tokens/nep171/mod.rs
```diff
@@ -31,6 +31,8 @@ async fn transfer_nft_to_verifier(mut rng: impl Rng) {
         .unwrap()
         .unwrap();
 
+    let persistent_tokens = user1.mt_tokens(env.defuse.id(), ..).await.unwrap();
```
fusede:
```suggestion
    let existing_tokens = user1.mt_tokens(env.defuse.id(), ..).await.unwrap();
```

tests/src/tests/defuse/env/builder.rs
```diff
@@ -126,27 +128,22 @@ impl EnvBuilder {
...
             defuse,
             wnear,
             poa_factory: poa_factory.clone(),
             sandbox,
             disable_ft_storage_deposit: self.disable_ft_storage_deposit,
             disable_registration: self.disable_registration,
-            persistent_state: None,
-            current_user_index: Mutex::new(0),
+            seed: Seed::from_entropy(),
+            new_user_index: AtomicUsize::new(0),
```
fusede:
```suggestion
            next_user_index: AtomicUsize::new(0),
```

tests/src/tests/defuse/env/mod.rs
```diff
@@ -147,34 +150,37 @@ impl Env {
...
+        } else {
+            generate_random_account_id(root, &mut Unstructured::new(&rand.random::<[u8; 64]>()))
         }
-
-        generate_random_account_id(self.sandbox.root_account().id())
     }
 
-    pub async fn upgrade_legacy(&mut self) {
-        self.generate_storage_data().await;
+    pub async fn upgrade_legacy(&self, create_unique_users: bool) {
```
fusede:
```suggestion
    pub async fn upgrade_legacy(&self, reuse_accounts: bool) {
```

tests/src/tests/defuse/env/storage.rs
```diff
@@ -29,58 +32,41 @@ use crate::{
 };
 
 impl Env {
-    pub async fn generate_storage_data(&mut self) {
-        let state =
-            PersistentState::generate(self.sandbox.root_account(), self.poa_factory.as_account());
-
-        self.persistent_state = Some(state);
+    pub async fn generate_storage_data(&self) -> Result<PersistentState> {
```
fusede:
```suggestion
    async fn generate_storage_data(&self) -> Result<PersistentState> {
```

tests/src/tests/defuse/env/storage.rs
```diff
@@ -29,58 +32,41 @@ use crate::{
...
-        futures::join!(self.apply_tokens(), self.apply_accounts());
+        self.apply_tokens(&state).await?;
+        self.apply_accounts(&state).await?;
 
-        self.apply_token_balances().await;
+        Ok(state)
     }
 
-    pub async fn verify_storage_consistency(&self) {
+    pub async fn verify_storage_consistency(&self, state: &PersistentState) {
```
fusede:
```suggestion
    async fn verify_storage_consistency(&self, state: &PersistentState) {
```

tests/src/tests/defuse/env/storage.rs
```diff
@@ -29,58 +32,41 @@ use crate::{
...
-        try_join_all(
-            state
-                .tokens
-                .iter()
-                .map(|token_id| self.apply_token(token_id)),
-        )
-        .await
-        .expect("Failed to apply tokens");
-    }
+    pub async fn apply_tokens(&self, state: &PersistentState) -> Result<()> {
```
fusede:
```suggestion
    async fn apply_tokens(&self, state: &PersistentState) -> Result<()> {
```

tests/src/tests/defuse/env/storage.rs
```diff
@@ -28,11 +29,37 @@ use crate::{
         },
         poa::factory::PoAFactoryExt,
     },
-    utils::{ParentAccount, mt::MtExt},
+    utils::{ParentAccount, acl::AclExt, mt::MtExt},
 };
 
 impl Env {
-    pub async fn generate_storage_data(&self) -> Result<PersistentState> {
+    pub async fn upgrade_legacy(&self, reuse_accounts: bool) {
```
fusede:
```suggestion
    async fn upgrade_legacy(&self, reuse_accounts: bool) {
```

PR #140, opened 2025-10-24
Salt generation fix

hlgltvnnk:
Salt generation fix: not only invalid salts but all salts are checked when new salt created

## Release Notes

* **New Features**
  * Added versioned nonce format parsing with backward compatibility for legacy nonce formats.

* **Improvements**
  * Refined salt rotation logic to ensure reliable and consistent salt selection based on usage history.
  * Enhanced salt state tracking during rotation to prevent premature salt reuse.

### DIFF
--- a/core/src/nonce/salted.rs
+++ b/core/src/nonce/salted.rs
@@ -152,7 +152,7 @@ impl SaltRegistry {
     }
 
     #[inline]
-    pub fn is_used(&self, salt: Salt) -> bool {
+    fn is_used(&self, salt: Salt) -> bool {
         salt == self.current || self.previous.get(&salt).is_some()
     }
 


core/src/nonce/salted.rs
```diff
@@ -256,4 +261,17 @@ mod tests {
...
     }
+
+    #[rstest]
+    fn derive_next_test(random_bytes: Vec<u8>) {
+        let mut salt_registry = SaltRegistry::new(random_bytes);
+
+        let prev = salt_registry.set_new().unwrap();
+
+        salt_registry.invalidate(prev).unwrap();
+        salt_registry.set_new().unwrap();
```
fusede: Why do you need `.set_new()` here?

PR #145, opened 2025-10-27
Remove randomness from tests

pityjllk:
- replaces random values with hardcoded dummy ones
- introduces fuzzy tests using https://github.com/proptest-rs/proptest
- introduces helper fn `create_defuse_payload` that automatically generates unique salted nonces
- dedicated test case for executing intent with legacy nonce

## Summary by CodeRabbit

* **Refactor**
  * Tests migrated from RNG/fixture-driven flows to deterministic, payload-driven and property-based tests for broader, more reliable coverage.

* **New Features**
  * Payload helpers and extensions: default payload signing, nonce-extraction trait, deterministic key fixtures, legacy-nonce test support, and a dummy message address constant.
  * Env helper to obtain unique nonces.
  * Public logging type re-exported and improved pretty-printed test logs.

* **Chores**
  * Enabled workspace property-testing tooling and added an "arbitrary" feature for tests.

### DIFF
--- a/tests/src/tests/defuse/tokens/nep245/mt_transfer_resolve_gas.rs
+++ b/tests/src/tests/defuse/tokens/nep245/mt_transfer_resolve_gas.rs
@@ -157,6 +157,15 @@ async fn run_resolve_gas_test(
     // );
 
     let non_existent_account = ArbitraryAccountId::arbitrary_as(&mut u).unwrap();
+    println!("Non-existent account: {non_existent_account}");
+
+    assert!(
+        env.defuse
+            .mt_tokens_for_owner(env.defuse.id(), &non_existent_account, ..=2) // 2 because we only need to check the first N tokens. Good enough.
+            .await
+            .unwrap()
+            .is_empty(),
+    );
 
     // We attempt to do a transfer of fictitious token ids from defuse to an arbitrary user.
     // These will fail, but there should be enough gas to do refunds successfully.


tests/src/tests/defuse/upgrade.rs
```diff
@@ -206,16 +179,12 @@ async fn test_upgrade_with_persistence(mut rng: impl Rng, random_bytes: Vec<u8>)
...
-            .clone()
-            .into_iter()
-            .chain(std::iter::once(Token {
-                token_id: TokenId::Nep141(Nep141TokenId::new(ft1.clone())).to_string(),
-                owner_id: None,
-            }))
-            .collect();
-
-        assert_eq!(tokens, expected);
+        persistent_tokens.push(Token {
```
hlgltvnnk: this is a previous version of the code - it was changed after the last review of the MR with tests

tests/src/tests/defuse/mod.rs
```diff
@@ -98,6 +108,84 @@ impl DefuseExt for Contract {
...
+    #[must_use]
+    async fn get_unique_nonce(
+        &self,
+        defuse_contract_id: &AccountId,
+        deadline: Option<Deadline>,
+    ) -> anyhow::Result<(Nonce, Deadline)>;
+}
+
+impl DefuseNonceExt for near_workspaces::Account {
+    async fn get_unique_nonce(
```
fusede:
```suggestion
impl DefuseNonceExt for near_workspaces::Account {
    async fn gen_unique_nonce(
```

fusede: Why is  this an extension of an `near_workspaces::Account` if it uses no info about the account itself? Maybe make it use `self.id()` as `defuse_contract_id` and call just  as `env.defuse.gen_unique_nonce(None)`?

tests/src/utils/payload.rs
```diff
@@ -0,0 +1,14 @@
...
+    fn extract_nonce(&self) -> Result<Nonce, serde_json::Error>;
+}
+
+impl ExtractNonceExt for MultiPayload {
+    #[inline]
+    fn extract_nonce(&self) -> Result<Nonce, serde_json::Error> {
+        let DefusePayload::<DefuseIntents> { nonce, .. } = self.clone().extract_defuse_payload()?;
+        Ok(nonce)
+    }
+}
```
fusede: I don't like this pattern, nonce can be retrieved even before payload was signed. See comment with `DefusePayloadBuilder` above

tip191/src/lib.rs
```diff
@@ -82,101 +78,61 @@ mod tests {
...
+    //     "a4b319a82adfc43584e4537fec97a80516e16673db382cd91eba97abbab8ca56",
+    // )?;
+    // let signing_key = wallet.signer();
+    // let verifying_key = signing_key.verifying_key();
+    // let public_key = verifying_key.to_encoded_point(false);
+    // // Notice that we skip the first byte, 0x04
+    // println!("Public key: 0x{}", hex::encode(public_key.as_bytes()[1..]));
+
+    const REFERENCE_MESSAGE: &str = "Hello, TRON!";
+    const NOK_REFERENCE_MESSAGE: &str = "this is not TRON reference input message";
```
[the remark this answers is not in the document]
pityjllk: nok => not ok

tests/src/tests/defuse/env/state.rs
```diff
@@ -96,47 +62,49 @@ impl PersistentState {
...
+                (
+                    subaccount,
+                    AccountWithTokens {
+                        data: AccountData {
+                            public_keys,
+                            nonces,
+                        },
+                        tokens: account_tokens,
+                    },
+                )
```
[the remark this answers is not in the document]
pityjllk: must have accidentaly removed `.take()` on tokens interator ... so the idea was to have different acounts with different number of toknes and different amount, sth like:
- USER1 : `[TOKEN1 : 1000]`
- USER2 : `([TOKEN1 : 2000], [TOKEN2: 2000])`
- USER3 : `([TOKEN1 : 3000], [TOKEN2: 3000], [TOKEN3: 3000])`

etc let me know if that works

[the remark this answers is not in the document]
fusede: Ok, good, I think it would be random enough. Just don't forget to add `.take(random_number)` for every iterator here

tests/src/tests/defuse/mod.rs
```diff
@@ -110,69 +110,56 @@ impl DefuseExt for Contract {
...
-    ) -> anyhow::Result<(Nonce, Deadline)> {
+    ) -> anyhow::Result<Nonce> {
+        deadline.unwrap_or_else(|| Deadline::timeout(std::time::Duration::from_secs(120)));
+        let deadline =
+            deadline.unwrap_or_else(|| Deadline::timeout(std::time::Duration::from_secs(120)));
         let seed_value = GLOBAL_SEED_COUNTER.fetch_add(1, Ordering::Relaxed);
+        let salt = self
+            .current_salt(defuse_contract_id)
+            .await
+            .expect("can fetch salt");
```
fusede:
```suggestion
            .expect("can't fetch salt");
```

tests/src/tests/defuse/mod.rs
```diff
@@ -110,69 +110,56 @@ impl DefuseExt for Contract {
...
+
         let mut nonce_bytes = [0u8; 15];
         TestRng::from_entropy().fill_bytes(&mut nonce_bytes);
         nonce_bytes[..8].copy_from_slice(&seed_value.to_le_bytes());
```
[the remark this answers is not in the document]
fusede: Hm, when why not using `GLOBAL_SEED_COUNTER` only?

fusede (approved):

hlgltvnnk (approved):

PR #148, opened 2025-11-05
Feat/receiver call after transfer intent

hlgltvnnk:
- mt_on_transfer() celled on receiver contract after transfer intent (if msg passed)
- TransferEvent added
- Covered with tests

## Summary by CodeRabbit

* **New Features**
  - Transfers can include an optional on-transfer notification (message + optional per-token min gas); receivers are notified and transfer events now emit a compact TransferEvent payload.

* **Tests**
  - Added a multi-token receiver stub and extensive transfer tests covering accept, partial refund, malicious refund, panic, and large return.
  - Test utilities updated to use new stub artifacts and adjusted WASM asset paths.

* **Documentation**
  - Added docs for the multi-token receiver stub and test guidance.

* **Chores**
  - Build task and env vars added for the stub; package version bumped and changelog updated.

### DIFF
--- a/core/src/intents/tokens.rs
+++ b/core/src/intents/tokens.rs
@@ -30,7 +30,7 @@ pub struct NotifyOnTransfer {
 
     // Min gas per single call
     /// If it is not specified optional minimum required Near gas will be used for created Promise to succeed:
-    /// * `mt_on_transfer`:      minimum: 5 TGas, default: 15TGas
+    /// * `mt_on_transfer`:      minimum: 5TGas, default: 30TGas
     ///
     /// Remaining gas will be distributed evenly across all Function Call
     /// Promises created during execution of current receipt.


core/src/accounts.rs
```diff
@@ -71,3 +71,22 @@ pub struct SaltRotationEvent {
...
+    all(feature = "abi", not(target_arch = "wasm32")),
+    serde_as(schemars = true)
+)]
+#[cfg_attr(
+    not(all(feature = "abi", not(target_arch = "wasm32"))),
+    serde_as(schemars = false)
+)]
+#[near(serializers = [json])]
+#[derive(Debug, Clone)]
+pub struct TransferEvent<'a> {
```
pityjllk: the only reason for that new event is so it doesnt include new `msg` field?

multi-token-receiver-stub/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+                    sha256(i.to_string().as_bytes());
+                }
+            }
+            MTReceiverMode::ExceedLogLimit => {
+                for _ in 0..100 {
+                    log!(
+                        "NEAR Intents is a multichain transaction protocol where users specify what they want and let third parties compete to provide the best solution. This works for everything from token swaps to pizza delivery, creating a universal marketplace across crypto and traditional services."
+                    );
+                }
+            }
```
pityjllk: it seems that in both cases contrac call will simply fail so i guess you might just use `panic` here. the other interesting scenario is to trigger vulnerability as for nep141. 

i did 

```
            StubAction::MaliciousReturn => {
                PromiseOrValue::Value(vec![U128(0xffffffffffffffffffffffffffffffff); 250000])
            }

```

just to cover also scenario of malicious external contract that will fail promise resolution on defuse side

tests/src/tests/defuse/env/mod.rs
```diff
@@ -266,6 +269,14 @@ impl Env {
             .unwrap();
     }
 
+    pub async fn deploy_mt_receiver_stub(&self) -> Contract {
```
pityjllk: 💯

multi-token-receiver-stub/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn decode(msg: &str) -> Self {
+        serde_json::from_str(msg)
+            .unwrap_or_else(|err| panic!("failed to deserialize MTReceiverMode: {err}"))
+    }
+
+    #[cfg(test)]
+    fn encode(&self) -> String {
+        serde_json::to_string(self).expect("MTReceiverMode::encode serialization should succeed")
+    }
+}
```
pityjllk: i guess it should be possible to simply use `serializaers = [json]` thing

core/src/accounts.rs
hlgltvnnk: From the beginning, the dip4 transfer event contained a Transfer structure to which we have to add a message that must be serialized and processed during intent execution but also have to be skipped during event emission. This means that we cannot do this with the serde macro (with skip_deserializing), because it is impossible to pass this message in transfer intent tests, so I decided to create a separate event for it

PR #157, opened 2025-11-24
feat: escrow swap (nep616)

fusede:


pvolnov:
## Proposal

The contract currently uses a precision price format. This makes it inconvenient to work with tokens that have different decimal places.

## Improvement

Add two fields to the `Params` struct:

* `decimal_from: u8` — decimals of the source token
* `decimal_to: u8` — decimals of the destination token

## Benefits

1. Integer-only arithmetic without extra conversions
2. Compatible with any token decimals
3. Simpler and more transparent calculations
4. No precision loss from fixed-point scaling
5. Price reflects a direct ratio in native token units

## Example (Standard)

Before:

```json
{ "price": "1500000000000" }   // 1.5 with fixed 12-decimal precision
```

After:

```json
{
  "price": "1500000",
  "decimal_from": 6,
  "decimal_to": 6
}
```

Formula:
`src_amount = (dst_amount * price) / 10^decimal_from`

## Example (Very Small Price)

Suppose the actual price is:

```
0.0000000000000001212
```

Instead of scaling this into a fixed 12-decimal format (which would lose precision), it can be expressed as an integer ratio.

For example, if both tokens have 6 decimals:

```json
{
  "price": "1212",          // numerator of the ratio
  "decimal_from": 6,
  "decimal_to": 24          // chosen to reflect the scale needed for small values
}
```

This represents:

```
price = 1212 / 10^24 = 0.0000000000000001212
```

The ratio remains exact and fully integer-based.

## Notes

* Price represents:
  `(dst_amount * 10^decimal_to) / (src_amount * 10^decimal_from)`
* No floating-point arithmetic is used
* Supports extremely small or large ratios without precision loss

---

PR #181, opened 2025-12-05
feat: escrow-swap contract

mitinarseny:
## Summary by CodeRabbit

* **New Features**
  * Added an escrow-swap contract with deterministic deployment, partial fills, multi-token support (NEP-141/NEP-245), auth-call flows, fee handling, lifecycle events, and resolution callbacks.
  * Introduced UD128 decimal type with parsing, formatting, arithmetic, and fuzz-tested roundtrips.
  * Added promise utilities and a unified checked-numeric operations API.

* **Documentation**
  * Comprehensive Escrow-Swap README describing flows, params, events, and usage.

* **Tests**
  * New integration tests for escrow flows and fuzz targets for decimal parsing.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/escrow-swap/src/contract/fill.rs
+++ b/escrow-swap/src/contract/fill.rs
@@ -112,7 +112,13 @@ impl State {
             .and_maybe(send_fees)
             .then(
                 self.callback_resolve_transfers(None, Some(maker_dst))
-                    .es_return_value(maker_dst.refund_value(taker_dst_in - taker_dst_used)?),
+                    .es_return_value(
+                        maker_dst.refund_value(
+                            taker_dst_in
+                                .checked_sub(taker_dst_used)
+                                .ok_or(Error::IntegerOverflow)?,
+                        )?,
+                    ),
             )
             .into())
     }


decimal/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+/// Floating point unsigned decimal price, i.e. dst per 1 src
+/// always reduced (i.e. normalized)
+#[derive(near_sdk::NearSchema)]
+#[abi(json, borsh)]
+#[schemars(with = "String")]
+#[derive(
+    Clone, Copy, PartialEq, Eq, Hash, BorshSerialize, SerializeDisplay, DeserializeFromStr,
+)]
+#[borsh(crate = "::near_sdk::borsh")]
+pub struct UD128(u8, u128);
```
pityjllk: This name is strange to me

escrow-swap/src/contract/return_value.rs
```diff
@@ -0,0 +1,28 @@
...
+pub trait ReturnValueExt: Sized {
+    fn return_value(self, value: Vec<u8>) -> Self;
+}
+
+impl ReturnValueExt for Promise {
+    fn return_value(self, value: Vec<u8>) -> Self {
+        self.function_call_weight(
+            "return_value".to_string(),
+            value,
+            NearToken::from_yoctonear(0),
```
pityjllk: maybe put it into dedicated crate so we could use it in other contracts?

decimal/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+    pub const fn new(mut decimals: u8, mut digits: u128) -> Option<Self> {
+        if digits == 0 {
+            return Some(Self::ZERO);
+        }
+
+        while decimals > 0 && digits % 10 == 0 {
+            digits /= 10;
+            decimals -= 1;
+        }
+
```
pityjllk: so i figure out there should be some more efficient way to do it and it seems to be called "binary decomposition"

here is refernce impl 

```
        macro_rules! strip {
            ($power:expr, $count:expr) => {
                while decimals >= $count && digits % $power == 0 {
                    digits /= $power;
                    decimals -= $count;
                }
            };
        }
        strip!(10_000_000_000_000_000, 16);
        strip!(100_000_000, 8);
        strip!(10_000, 4);
        strip!(100, 2);
        strip!(10, 1);


```
ai explanation
```
  Insight: Any integer N can be represented as a sum of powers of 2 (its binary representation). Instead of stripping 1 zero at a time, we
  strip in chunks of 16, 8, 4, 2, 1.

  Example with N = 18:
  18 in binary = 10010 = 16 + 2

  Instead of: 18 operations (subtract 1, eighteen times)
  We do:      2 operations  (subtract 16, then subtract 2)

  Why powers of 2?

  Any integer 0-38 can be expressed using at most 6 powers of 2. This is the minimal representation — no other set of chunk sizes guarantees
   fewer pieces for all possible values.

  strip!(10^16, 16)  →  handles the 16-bit
  strip!(10^8,  8)   →  handles the 8-bit
  strip!(10^4,  4)   →  handles the 4-bit
  strip!(10^2,  2)   →  handles the 2-bit
  strip!(10^1,  1)   →  handles the 1-bit

  Complexity:
  - Original: O(N) where N = trailing zeros (up to 38)
  - Optimized: O(log N) — bounded to ~6 iterations max

  Same principle as: Exponentiation by squaring, binary search, and other divide-and-conquer algorithms that halve the problem space at each
   step.

```

decimal/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+    #[inline]
+    pub const fn digits(&self) -> u128 {
+        self.1
+    }
+
+    #[inline]
+    pub(crate) const fn denominator(self) -> u128 {
+        // this is safe since decimals are always <= Self::MAX_DECIMALS
+        #[allow(clippy::as_conversions)]
+        10u128.pow(self.decimals() as u32)
```
pityjllk: if we want to be even more efficient 🤔 

```
    const POW10: [u128; 39] = {
        let mut table = [1u128; 39];
        let mut i = 1usize;
        while i < 39 {
            table[i] = table[i - 1] * 10;
            i += 1;
        }
        table
    };


```

pityjllk (changes requested):

PR #193, opened 2026-01-12
feat: `imt_mint`/`imt_burn` intents

hlgltvnnk:
## Summary by CodeRabbit

* **New Features**
  * IMT support: multi-token mint/burn intents, new IMT token-id type/parsing, and IMT events.
* **Role & Admin**
  * Added UnrestrictedAccountManager role and bulk add/remove public-key management.
* **Tests**
  * Extensive IMT mint/burn scenarios and public-key synchronization tests added.
* **Chores**
  * Feature flags, build/CI updates, centralized token-id length validation/error, and minor gas tuning for native withdrawals.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/defuse/src/contract/tokens/nep141/native.rs
+++ b/defuse/src/contract/tokens/nep141/native.rs
@@ -5,7 +5,7 @@ use crate::contract::{Contract, ContractExt};
 
 #[near]
 impl Contract {
-    pub(crate) const DO_NATIVE_WITHDRAW_GAS: Gas = Gas::from_tgas(10);
+    pub(crate) const DO_NATIVE_WITHDRAW_GAS: Gas = Gas::from_tgas(12);
 
     #[private]
     pub fn do_native_withdraw(withdraw: NativeWithdraw) -> Promise {
--- a/tests/src/tests/defuse/intents/imt_mint.rs
+++ b/tests/src/tests/defuse/intents/imt_mint.rs
@@ -113,7 +113,7 @@ async fn failed_imt_mint_intent() {
 
     env.simulate_and_execute_intents(env.defuse.id(), [mint_payload.clone()])
         .await
-        .assert_err_contains("token ID is too large");
+        .assert_err_contains("token_id is too long");
 }
 
 #[rstest]


defuse/src/account_sync.rs
```diff
@@ -0,0 +1,17 @@
+use defuse_core::crypto::PublicKey;
+use near_sdk::{AccountId, ext_contract};
+
+// TODO: locked accounts + accounts with predecessor id disabled should be also synchronized
+#[ext_contract(ext_pk_sync_manager)]
+pub trait AccountSyncManager {
+    /// Registers or re-activates `public_key` under the user account_id.
+    ///
+    /// NOTE: MUST attach 1 yⓃ for security purposes.
+    fn force_add_public_keys(&mut self, public_keys: Vec<(AccountId, Vec<PublicKey>)>);
```
mitinarseny: Maybe avoid duplicate entries by design?
```suggestion
    fn force_add_public_keys(&mut self, public_keys: HashMap<AccountId, HashSet<PublicKey>>);
```

defuse/src/contract/mod.rs
```diff
@@ -53,6 +54,8 @@ pub enum Role {
     SaltManager,
 
     GarbageCollector,
+
+    PubKeySynchronizer,
```
mitinarseny: By force-adding a new public key the grantee of this role is available to fully control the account which key is being added to. Let's rename it to `UnrestrictedAccountManager` and let it call all methods which are allowed for `UnrestrictedWithdrawer` and `UnrestrictedAccountLocker/Unlocker`

core/src/engine/state/mod.rs
```diff
@@ -127,4 +127,14 @@ pub trait State: StateView {
...
+
+    fn mt_mint(&mut self, owner_id: AccountId, tokens: Amounts, memo: Option<String>)
+    -> Result<()>;
+
+    fn mt_burn(
+        &mut self,
+        owner_id: &AccountIdRef,
+        tokens: Amounts,
+        memo: Option<String>,
+    ) -> Result<()>;
```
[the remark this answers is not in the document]
mitinarseny: As for renaming: this is exactly why I want it to be renamed, since it gets signer_id from the caller

core/src/intents/tokens.rs
```diff
@@ -57,6 +60,16 @@ impl NotifyOnTransfer {
...
+
+    pub fn set_gas(mut self) -> Self {
+        self.min_gas = Some(
+            self.min_gas
+                .unwrap_or(Self::MT_ON_TRANSFER_GAS_DEFAULT)
+                .max(Self::MT_ON_TRANSFER_GAS_MIN),
+        );
+
+        self
+    }
```
mitinarseny: let's remove it from here and use it in `Contract::notify_on_transfer` method, so that callers won't have to worry about remembering to call `.set_gas()` before passing it there

core/src/intents/tokens.rs
```diff
@@ -561,51 +566,33 @@ impl ExecutableIntent for MtMint {
...
-        I: Inspector,
-    {
-        engine.inspector.on_event(DefuseEvent::MtBurn(Cow::Borrowed(
+        engine.inspector.on_event(DefuseEvent::MtMint(Cow::Borrowed(
             [IntentEvent::new(
-                AccountEvent::new(owner_id, Cow::Borrowed(&self)),
+                AccountEvent::new(
+                    owner_id,
+                    Cow::Owned(Self {
+                        tokens: tokens.clone(),
```
mitinarseny: mutate in-place afterwards

mitinarseny (changes requested):

core/src/intents/tokens.rs
```diff
@@ -556,43 +543,82 @@ impl ExecutableIntent for MtMint {
...
+                .as_slice(),
+            )));
+
+        engine
+            .state
+            .mt_burn(owner_id, self.tokens.to_imt_tokens(owner_id), self.memo)
+    }
+}
+
+pub type MintedTokens = Amounts<BTreeMap<defuse_nep245::TokenId, u128>>;
```
mitinarseny: What do you think if we rename `MintedTokens` to `ImtTokens`?

mitinarseny (changes requested):

core/src/intents/tokens.rs
```diff
@@ -556,43 +543,82 @@ impl ExecutableIntent for MtMint {
...
-                    Cow::Owned(Self {
-                        tokens: tokens.clone(),
-                        ..self.clone()
-                    }),
-                ),
-                intent_hash,
-            )]
-            .as_slice(),
-        )));
+        let tokens = self.tokens.to_imt_tokens(owner_id);
```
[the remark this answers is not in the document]
hlgltvnnk: I dont think there will be a need for this, since we dont check this in mt_on_transfer - I dont see any malicious ways to use this, with large amounts of tokens, the user will either run into logs limit. Although, if we still want to fight spam, this will need to be implemented

core/src/intents/tokens.rs
mitinarseny: That's why I'm suggesting to rename it. These are imt-specific tokens, not a generic ones yet

core/src/intents/tokens.rs
```diff
@@ -556,43 +543,82 @@ impl ExecutableIntent for MtMint {
...
+        engine
+            .state
+            .mt_burn(owner_id, self.tokens.to_imt_tokens(owner_id), self.memo)
+    }
+}
+
+pub type MintedTokens = Amounts<BTreeMap<defuse_nep245::TokenId, u128>>;
+
+impl MintedTokens {
+    fn to_imt_tokens(&self, minter_id: &AccountIdRef) -> Amounts<BTreeMap<TokenId, u128>> {
```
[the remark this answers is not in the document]
mitinarseny: The conversion goes from imt-specific to generic ones

defuse/src/contract/tokens/mod.rs
```diff
@@ -53,7 +53,7 @@ impl Contract {
                         return Err(DefuseError::NftAlreadyDeposited(tid.clone()));
                     }
                 }
-                TokenId::Nep141(_) | TokenId::Nep245(_) => {}
+                TokenId::Nep141(_) | TokenId::Nep245(_) | TokenId::Imt(_) => {}
```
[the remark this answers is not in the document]
mitinarseny: It's still TBD whether we should enable this feature for both Near and Far. Let's put it behind `imt` feature, so we have an option to disable it on `intents.near`, while having it enabled on `intents.far`

core/src/intents/tokens.rs
```diff
@@ -573,6 +577,10 @@ impl ExecutableIntent for ImtMint {
 #[derive(Debug, Clone)]
 /// Burn a set of tokens minted by signer, within the intents contract.
 pub struct ImtBurn {
+    // The tokens transferred in this call will be wrapped
+    // in such a way as to bind the token ID to the minter authority.
+    // The final string representation of the token
+    // will be as follows: imt:<minter_id>:<token_id>
```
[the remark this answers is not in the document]
mitinarseny: only minter can mint, anyone can burn

PR #202, opened 2026-02-02
feat: add state_init benchmarks and fix gas accounting

pityjllk:
- Added benchmarks that validate STATE_INIT_GAS constant is sufficient when deploying contracts as part of transfer notification or auth call intents
  - Moved STATE_INIT_GAS constant to dedicated consts module and removed the TODO comment
  - Fixed gas accounting for transfer notifications with state_init by adding extra gas overhead
  - Implemented on_auth callback in multi-token-receiver-stub test contract for auth call testing
  - Added sandbox extensions for deploying global contracts and their instances
  - Benchmarks confirm 10 Tgas is sufficient for state_init with payloads up to ~560 bytes (zero-balance account limit)
  - Bumped near-api dependency to 0.8.3

  **STATE_INIT_GAS covers the cost of deploying a deterministic account via StateInit (NEP-616) when the account doesn't exist yet. It only accounts for deployments via Global Contract reference (NEP-591) with less than 770
  bytes of storage which doesn't require storage staking. For larger deployments that need more gas, callers should use AuthCall::min_gas or NotifyOnTransfer::min_gas and provide storage deposit separately.**

## Summary by CodeRabbit

* **New Features**
  * MT receiver stub deployment utilities, deterministic state-init in transactions, raw intent execution API, and a deploy-global-contract helper.

* **Documentation**
  * Gas constant made public and increased; added a public helper to compute auth-call callback gas with explanatory comments.

* **Library**
  * Added Remainder adapter and improved multi-branch borsh deserialization behavior.

* **Tests**
  * New integration and benchmark suites for global deployment, state-init, escrow, and auth-call scenarios.

* **Chores**
  * Dependency bumps, sandbox feature/config updates, and audit ignore entry added.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -7,4 +7,4 @@
 # The advisory relates to IterMut's Stacked Borrows violation, which does
 # not affect our usage as we don't use IterMut directly.
 # Tracked: https://github.com/near/nearcore/issues/XXXXX (upstream)
-ignore = ["RUSTSEC-2026-0002"]
+ignore = ["RUSTSEC-2026-0002", "RUSTSEC-2026-0009"]


mitinarseny (changes requested):
It seems that in your tests you're checking for burnt gas on the receipt that had `StateInit + FunctionCall` actions, while the main goal is to check if [`do_auth_call` method](https://github.com/near/intents/blob/9e45cccb322744c34ea0265d7850ff12d012f39f/defuse/src/contract/intents/auth_call.rs#L13-L14) gets [enough gas](https://github.com/near/intents/blob/9e45cccb322744c34ea0265d7850ff12d012f39f/defuse/src/contract/intents/state.rs#L327-L340) when `auth_call` intent is executed with non-zero [attached_deposit](https://github.com/near/intents/blob/9e45cccb322744c34ea0265d7850ff12d012f39f/core/src/intents/auth.rs#L27-L33). The latter is needed to trigger `do_auth_call` to be executed in a separate receipt.

tests/src/tests/defuse/state_init.rs
```diff
@@ -173,90 +135,103 @@ async fn benchmark_auth_call_with_state_init(mut rng: impl Rng) {
...
-    let (user, ft) = futures::join!(env.create_user(), env.create_token());
-
-    env.initial_ft_storage_deposit(vec![user.id()], vec![ft.id()])
-        .await;
+async fn benchmark_gas_used_by_do_auth_call_callback(mut rng: impl Rng) {
+    // NOTE: when do_auth_call is scheduled as callback to withdraw (because of
+    // AuthCall::storage_deposit > 0) it needs to check status of withdrawal. We can't trigger
+    // it in this case so we need to subtract gas for promise read (it's around 0.1Tgas) with
+    // some overhead.
+    const NEAR_WITHDRAW_PROMISE_READ_OVERHEAD: Gas = Gas::from_tgas(1);
```
mitinarseny: you can trigger it by having non-zero `auth_call.attached_deposit`

tests/src/tests/defuse/state_init.rs
```diff
@@ -173,90 +135,103 @@ async fn benchmark_auth_call_with_state_init(mut rng: impl Rng) {
...
+                fees: FeesConfig {
+                    fee: Pips::ZERO,
+                    fee_collector: env.id().clone(),
+                },
+                roles: RolesConfig::default(),
+            },
+            crate::env::DEFUSE_WASM.clone(),
+        )
+        .await
+        .unwrap();
```
mitinarseny: verifier contract  doesn't have `on_auth` method, why do you need it?

mitinarseny: I mean, you already have the `env.defuse`, right?

defuse/src/contract/consts.rs
```diff
@@ -0,0 +1,7 @@
+use near_sdk::Gas;
+
+// Covers StateInit (NEP-616) cost when deterministic account doesn't exist yet.
+// Only accounts for deploying via Global Contract ref (NEP-591) with <770B storage
+// which doesn't require storage staking. If you need to attach more GAS, utilize
+// `AuthCall::min_gas` or `NotifyOnTransfer::min_gas` and provide storage deposit separately.
```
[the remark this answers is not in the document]
pityjllk: removed

mitinarseny (changes requested):

tests/contracts/multi-token-receiver-stub/src/lib.rs
```diff
@@ -39,10 +39,47 @@ impl ReturnValueExt for Promise {
...
+}
+
+impl near_sdk::borsh::BorshDeserialize for AnyData {
+    fn deserialize_reader<R: std::io::Read>(reader: &mut R) -> std::io::Result<Self> {
+        // Consume and ignore all remaining bytes
+        let mut buf = Vec::new();
+        reader.read_to_end(&mut buf)?;
+        Ok(AnyData)
+    }
+}
```
mitinarseny: A suggestion to make it more re-usable:
```rust
pub struct Remainder;

impl<T> BorshSerializeAs<T> for Remainder
where
    T: AsRef<[u8]>
{
    fn serialize<W: io::Write>(source: T, writer: &mut W) -> io::Result<()> {
        writer.write_all(source.as_ref())
    }
}

impl<T> BorshDeserializeAs<T> for Remainder
where
    T: TryFrom<Vec<u8>>,
    T::Error: Into<Box<dyn Error + Send + Sync>>,
{
    fn deserialize_reader<R: io::Read>(reader: &mut R) -> io::Result<T> {
        let mut buf = Vec::new();
        reader.read_to_end(&mut buf)?;
        buf.try_into().map_err(|err| io::Error::new(io::ErrorKind::InvalidData, err))
    }
}
```

Put it in `defuse-borsh-utils`. Then, it can be used as:
```rust
#[derive(BorshSerialize, BorshDeserialize)]
struct S {
    #[borsh(
        serialize_with = "As::<Remainder>::serialize",
        deserialize_with = "As::<Remainder>::deserialize"
    )]
    pub a: Vec<u8>,
}
```

pityjllk: https://github.com/near/intents/blob/3afc15c2c1cbfedd9ab700b04ea39bdc51c764f1/tests/contracts/multi-token-receiver-stub/src/lib.rs#L43-L53

done

mitinarseny (approved):

PR #205, opened 2026-02-05
feat: multifunctional events

hlgltvnnk:
## Summary by CodeRabbit

* **New Features**
  * Unified, structured event system for account actions, transfers, and IMT operations; standardized transfer and mint event payloads (including receiver, token amounts, optional memo).

* **Refactor**
  * Token-related constants and types moved to a public tokens surface for easier reuse.
  * Events now support intent-wrapping (including intent hashes) to improve event traceability and wrapping of account events.

* **Tests**
  * Added comprehensive backward-compatibility tests for event serialization and formats.

### DIFF
--- a/core/src/intents/mod.rs
+++ b/core/src/intents/mod.rs
@@ -163,7 +163,7 @@ pub struct MaybeIntentEvent<T> {
 
 impl<T> MaybeIntentEvent<T> {
     #[inline]
-    pub const fn new(event: T) -> Self {
+    pub const fn new_fn_call(event: T) -> Self {
         Self {
             intent_hash: None,
             event,
@@ -171,7 +171,7 @@ impl<T> MaybeIntentEvent<T> {
     }
 
     #[inline]
-    pub const fn new_with_hash(event: T, intent_hash: CryptoHash) -> Self {
+    pub const fn new_intent(event: T, intent_hash: CryptoHash) -> Self {
         Self {
             intent_hash: Some(intent_hash),
             event,
@@ -181,7 +181,7 @@ impl<T> MaybeIntentEvent<T> {
 
 impl<T> From<T> for MaybeIntentEvent<T> {
     fn from(event: T) -> Self {
-        Self::new(event)
+        Self::new_fn_call(event)
     }
 }
 
--- a/defuse/src/contract/accounts/mod.rs
+++ b/defuse/src/contract/accounts/mod.rs
@@ -97,7 +97,7 @@ impl Contract {
         let toggled = self.internal_set_auth_by_predecessor_id(account_id, enable, force)?;
 
         if toggled {
-            DefuseEvent::SetAuthByPredecessorId(MaybeIntentEvent::new(AccountEvent::new(
+            DefuseEvent::SetAuthByPredecessorId(MaybeIntentEvent::new_fn_call(AccountEvent::new(
                 Cow::Borrowed(account_id),
                 Cow::Owned(SetAuthByPredecessorId { enabled: enable }),
             )))
@@ -137,7 +137,7 @@ impl Contract {
     ) {
         State::add_public_key(self, account_id.into(), public_key).unwrap_or_panic();
 
-        DefuseEvent::PublicKeyAdded(MaybeIntentEvent::new(AccountEvent::new(
+        DefuseEvent::PublicKeyAdded(MaybeIntentEvent::new_fn_call(AccountEvent::new(
             Cow::Borrowed(account_id),
             PublicKeyEvent {
                 public_key: Cow::Borrowed(&public_key),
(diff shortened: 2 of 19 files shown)


pityjllk (changes requested):
it feels like we could get rid of the one of the wrapping event types 🤔

core/src/events.rs
```diff
@@ -1,32 +1,61 @@
...
 
 #[cfg(feature = "imt")]
-use crate::intents::tokens::imt::{ImtBurn, ImtMint};
+use crate::{intents::tokens::imt::ImtBurn, tokens::imt::ImtMintEvent};
+
+#[must_use]
+#[near(serializers = [json])]
+#[serde(untagged)]
+#[derive(Debug, Clone)]
+pub enum ContractEvent<T> {
```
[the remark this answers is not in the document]
hlgltvnnk: This was done for events that can be emitted simultaneously in intents and in fn calls. Using two types lets us enforce this at the type level and keep event schemas explicit

sandbox/src/extensions/defuse/event.rs
```diff
@@ -0,0 +1,345 @@
...
+                    .flat_map(move |i| i.into_defuse_events(signer_id.clone(), hash))
+            })
+            .collect();
+
+        if !executed.is_empty() {
+            events.push(DefuseEvent::IntentsExecuted(executed.into()));
+        }
+
+        events
+    }
```
pityjllk: 1. replace unwrap with unreachable
2. maybe rework whole thing to collect events to Vec<(MaybeIntentEvent, Vec<DefuseEvents>>. then `unzip` it to Vec<MaybeIntentEvent> and Vec<Vec<DefuseEvents>> . and then flatten ` Vec<Vec<DefuseEvents>>` and chain with whats held in `Vec<MaybeIntentEvent>`. 

main pain point here is that you have iterator loop that collects stuff and also pushes element to other collection. you could also omit `mut` modifier

sandbox/src/extensions/defuse/force_manager.rs
```diff
@@ -37,43 +38,49 @@ impl ForceAccountManagerExt for SigningAccount {
         &self,
         contract_id: impl Into<AccountId>,
         account_id: impl AsRef<AccountIdRef>,
-    ) -> anyhow::Result<bool> {
-        self.tx(contract_id)
+    ) -> anyhow::Result<(ExecutionSuccess, bool)> {
```
pityjllk: 💯

pityjllk (changes requested):

defuse/src/contract/accounts/mod.rs
```diff
@@ -79,19 +85,55 @@ impl Contract {
         predecessor_account_id
     }
 
+    pub fn set_auth_by_predecessor_id_and_emit_event(
+        &mut self,
+        account_id: &AccountId,
+        enable: bool,
+        force: bool,
+    ) -> Result<bool> {
+        DefuseEvent::SetAuthByPredecessorId(MaybeIntentEvent::new(AccountEvent::new(
```
pityjllk: i think now its emitted unconditionally as opposed to previous version 🤔 

https://github.com/near/intents/blob/main/defuse/src/contract/accounts/account/mod.rs#L139-L155

hlgltvnnk: Previously, the event was emitted in the middle of internal contract implementstion — which were called both with a regular fn call and with intent - which meant that in the case of intent, the event was emitted twice

core/src/intents/mod.rs
```diff
@@ -147,20 +147,44 @@ impl ExecutableIntent for Intent {
...
+        }
+    }
+
+    #[inline]
+    pub const fn new_with_hash(event: T, intent_hash: CryptoHash) -> Self {
+        Self {
+            intent_hash: Some(intent_hash),
+            event,
+        }
     }
```
mitinarseny: Maybe it's worth forcing devs to **always** make it explicit and think whether they should pass `None` or `intent_hash`? And remove `impl From<T> for MaybeIntentEvent<T>` at all?
```suggestion
    pub fn new(event: T, intent_hash: impl Into<Option<CryptoHash>> ) -> Self {
        Self {
            intent_hash: intent_hash.into(),
            event,
        }
    }
```
Wdyt?

defuse/src/contract/accounts/mod.rs
```diff
@@ -79,19 +85,59 @@ impl Contract {
...
+        &mut self,
+        account_id: &AccountIdRef,
+        enable: bool,
+        force: bool,
+    ) -> Result<bool> {
+        if enable {
+            let Some(account) = self.accounts.get_mut(account_id) else {
+                // no need to create an account: not-yet-existing accounts
+                // have auth by PREDECESSOR_ID enabled by default
+                return Ok(true);
```
mitinarseny: This will make it always emit `SetAuthByPredecessorId` for not-yet-existing accounts

```suggestion
    pub(crate) fn internal_set_auth_by_predecessor_id(
        &mut self,
        account_id: &AccountIdRef,
        enable: bool,
        force: bool,
    ) -> Result<bool> {
        if enable {
            let Some(account) = self.accounts.get_mut(account_id) else {
                // no need to create an account: not-yet-existing accounts
                // have auth by PREDECESSOR_ID enabled by default
                return Ok(false);
```

mitinarseny: Add docs for the returning value of this method

core/src/intents/mod.rs
hlgltvnnk: I would leave the two constructor methods for clarity, but I would revert the naming to the previous ones -  new_direct and new_intent - so that there would be no confusion about whether to pass than option or not, and it would reduce the amount of unnecessary boilerplate with options

PR #210, opened 2026-02-16
feat: checked typed api for promise results

pityjllk:
## Summary by CodeRabbit

* **Refactor**
  * Unified cross-contract result handling across token, NFT, MT and escrow flows by switching to new promise-result helpers; internal logic simplified and serde_json usage reduced, no public API changes.

* **New Features**
  * Added utilities to check void promise results, decode JSON promise results with length guards, and estimate max JSON payload lengths.

* **Tests**
  * Added tests for JSON length estimation and promise-result decoding behavior.


escrow-swap/src/contract/tokens/mod.rs
```diff
@@ -152,46 +153,37 @@ impl Sent {
...
+            // Do not refund on failure due to vulnerability in reference
+            // implementations of FT and MT: `{ft,mt}_resolve_transfer`
+            // can fail to read huge result returned by `{ft,mt}_on_transfer`
+            // due to insufficient gas
             match self.token_type {
                 #[cfg(feature = "nep141")]
                 TokenIdType::Nep141 => {
                     // `ft_transfer_call` returns successfully transferred amount
-                    serde_json::from_slice::<U128>(&value).unwrap_or_default().0
+                    match promise_result_checked_json::<U128>(result_idx) {
```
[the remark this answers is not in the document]
hlgltvnnk: I mean processing the results at *_transfer and *_transfer_call, so it can be shared betweed defuse and escrow

hlgltvnnk (dismissed):
lgtm

escrow-swap/src/contract/tokens/mod.rs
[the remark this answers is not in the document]
pityjllk: i think there are bits that are not exactly alligned, like in case of nep-245 escrow expects exactly single token etc ...

near-utils/src/promise.rs
```diff
@@ -10,3 +10,126 @@ impl PromiseExt for Promise {
...
+    fn max_json_length(_args: ()) -> usize {
+        " \"\" ".len() + "+340282366920938463463374607431768211455".len()
+    }
+}
+
+impl<T> MaxJsonLength for Vec<T>
+where
+    T: MaxJsonLength<Args = ()>,
+{
+    type Args = usize;
```
mitinarseny: Make it more generic:
```suggestion
impl<T> MaxJsonLength for Vec<T>
where
    T: MaxJsonLength,
{
    type Args = (usize, T::Args);
```

mitinarseny: Also, please add impl for `[T; N]` and use it in escrow-swap resolving logic

near-utils/src/promise.rs
```diff
@@ -10,3 +10,126 @@ impl PromiseExt for Promise {
...
+/// (e.g. `ft_transfer`, `nft_transfer`, `mt_batch_transfer`).
+#[inline]
+pub fn promise_result_checked_void(result_idx: u64) -> PromiseResult<()> {
+    let data = env::promise_result_checked(result_idx, 0)?;
+    if data.is_empty() {
+        Ok(())
+    } else {
+        unreachable!()
+    }
+}
```
mitinarseny: It seems redundant, since `env::promise_result_checked(0, 0)` is doing the same

near-utils/src/promise.rs
```diff
@@ -10,3 +10,126 @@ impl PromiseExt for Promise {
...
     }
 }
+
+pub type PromiseResult<T> = Result<T, near_sdk::PromiseError>;
+pub type PromiseJsonResult<T> = Result<Result<T, serde_json::Error>, near_sdk::PromiseError>;
+
+pub trait MaxJsonLength: DeserializeOwned {
+    type Args;
+    fn max_json_length(args: Self::Args) -> usize;
+}
```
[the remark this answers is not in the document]
mitinarseny: Recursive types implementing this trait would hit stack overflow:
```rust
pub struct S {
    pub list: Vec<S>,
}

impl MaxJsonLength for S {
    type Args = usize; // length

    fn max_json_length(length: usize) -> usize {
        "{\n        "list": \n}".len() + Vec::<S>::max_json_length();
    }
}
```

However, this can be solved using `depth` arg from [this comment](https://github.com/near/intents/pull/210/changes#r2815800198): you can define `pub const MAX_JSON_LENGTH_RECURSION_LIMIT: usize = 100` and check for it in all impls that bump `depth` by one before passing it to the underlying type:

```rust
const MAX_JSON_LENGTH_RECURSION_LIMIT: usize = 1024;

impl<T> MaxJsonLength for Vec<T>
where
    T: MaxJsonLength,
{
    /// length
    type Args = (usize, T::Args);

    fn max_json_length(depth: usize, (length, args): Self::Args) -> usize {
        if depth >= MAX_JSON_LENGTH_RECURSION_LIMIT {
            return usize::MAX;
        }

        let ident = "        ".len().saturating_mul(depth + 1);

        ident
            .saturating_add(T::max_json_length(depth + 1, args))
            .saturating_add(",\n".len())
            .saturating_mul(length)
            .saturating_add(" [\n ".len() + ident + "]".len())
    }
}
```

mitinarseny (changes requested):

near-utils/src/promise.rs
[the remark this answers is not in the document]
pityjllk: > so people would not need to import a lot of different functions:

we just have 2 functions right? one for json formatted result and another for empty case. theoretically there could be some specialization for `promise_result_checked_json` for `()` but empty json is not technically `json` thats why i went with dedicated fn...

near-utils/src/promise.rs
mitinarseny: Ok, then let's maybe change it to assert in debug-build only?

```suggestion
/// Returns `Ok(())` if the promise at `result_idx` succeeded with an empty result.
/// This is the expected outcome for void-returning cross-contract calls
/// (e.g. `ft_transfer`, `nft_transfer`, `mt_batch_transfer`).
#[inline]
pub fn promise_result_checked_void(result_idx: u64) -> PromiseResult<()> {
    let data = env::promise_result_checked(result_idx, 0)?;
    debug_assert!(data.is_empty());
    Ok(())
}
```

near-utils/src/promise.rs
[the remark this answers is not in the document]
pityjllk: added another helper to mittigate from ugly signatures

``` 
promise_result_checked_json_with_len
```

should be good enough for some time so in contract code we use 


``` 
        promise_result_checked_json_with_len::<Vec<U128>>(0, tokens_count)
        promise_result_checked_void(0)
        promise_result_checked_json::<U128>(result_idx)
        promise_result_checked_json::<[U128; 1]>(result_idx)
```

near-utils/src/promise.rs
```diff
@@ -10,3 +10,191 @@ impl PromiseExt for Promise {
         if let Some(p) = p { self.and(p) } else { self }
     }
 }
+
+pub type PromiseResult<T> = Result<T, near_sdk::PromiseError>;
+pub type PromiseJsonResult<T> = Result<Result<T, serde_json::Error>, near_sdk::PromiseError>;
+
+pub const MAX_JSON_LENGTH_RECURSION_LIMIT: usize = 1024;
```
mitinarseny: Can you pls double-check what are recursion limits on wasm32 or does it depend on particular WASM runtime being used?

mitinarseny (commented):
LGTM, pls address few more comments and we're good to go

near-utils/src/promise.rs
pityjllk: so ther no recursion limit per se. whats limited is stack size and its currently set to 16384. also the whole interface is generic so its hard to come up with single value that will work in all cases. cause different impl can account for different amount of operations, arguments etc. taking that into account lets limit it to some reasnably low number like `32` ?

mitinarseny (approved):

PR #222, opened 2026-03-04
fix: unrestricted manager role under far feature

hlgltvnnk:
## Summary by CodeRabbit

* **New Features**
  * Adds an optional "far" feature that enables extended account-management operations and an additional administrative role; API parity preserved when the feature is disabled via a hidden placeholder.

* **Chores**
  * Updated build, test, and configuration to reference and opt into the new "far" feature so the conditional capabilities are enabled at build time.

### DIFF
--- a/test-utils/src/wasms.rs
+++ b/test-utils/src/wasms.rs
@@ -33,8 +33,8 @@ pub fn read_wasm(mode: &ReadWasmMode, path: impl AsRef<Path>) -> Vec<u8> {
 }
 
 #[cfg(feature = "defuse")]
-pub static DEFUSE_IMT_WASM: LazyLock<Vec<u8>> =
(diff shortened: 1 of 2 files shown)


defuse/src/contract/mod.rs
```diff
@@ -54,6 +54,7 @@ pub enum Role {
 
     GarbageCollector,
 
+    #[cfg(feature = "far")]
     UnrestrictedAccountManager,
```
mitinarseny: We need to reserve this role bit even if `far` feature is not enabled
For example:
```suggestion
    #[cfg(feature = "far")]
    UnrestrictedAccountManager,
    #[cfg(not(feature = "far"))]
    #[deprecated]
    __UnrestrictedAccountManager,
```

mitinarseny (approved):

PR #237, opened 2026-03-17
feat:cargo check features

pityjllk:
Summary

  We use cargo-hack to verify that every feature in the workspace compiles correctly in isolation — both on the host and for wasm32-unknown-unknown. This catches issues like missing feature
  gates, unused imports behind cfg flags, and broken feature combinations that normal cargo check misses because it only tests default features.

  The check runs individual cargo check invocations across the workspace, covering:
  - Each feature tested alone (--each-feature) for most crates
  - Full powerset of variant features (--feature-powerset --at-least-one-of) for crates where enums are entirely feature-gated (defuse-token-id, defuse-ton-connect, defuse-escrow-swap)
  - Warnings treated as errors (-D warnings) on all passes

  Changes

  - Makefile: Added check-all-features-host, check-all-features-wasm, and check-examples targets
  - CI: Installs cargo-hack and runs make check-all-features on every PR
  - Cargo.toml: Enabled arbitrary/derive at workspace level (near-sdk uses derive(Arbitrary) without enabling the feature itself)
  - Feature fixes: Added missing feature gates for defuse-sep53 (ed25519), defuse-ton-connect (abi propagation), defuse-token-id (arbitrary implies nep245), and conditional imports in
  ton-connect/schema/mod.rs
  - sandbox: Gated imt module behind its feature flag

## Summary by CodeRabbit

* **Chores**
  * CI now runs stricter WASM/clippy checks plus workspace-wide feature and example checks.
  * Build tooling expanded to exercise all feature combinations and host/wasm variants; help output updated.
  * Formatting flow extended to include additional formatters.
  * Adjusted workspace and crate dependency feature declarations for compatibility.

* **Refactor**
  * Public API signatures updated for consistency.
  * Added conditional feature-gating to modules and methods to reduce compilation surface.
  * Updated cryptography feature selections in signature crates.


mitinarseny (commented):
wow

hlgltvnnk (dismissed):

.github/workflows/ci.yml
```diff
@@ -52,11 +52,17 @@ jobs:
...
       - name: Install cargo-machete
         run: curl -fsSL https://github.com/bnjbvr/cargo-machete/releases/download/v0.9.1/cargo-machete-v0.9.1-x86_64-unknown-linux-musl.tar.gz | tar xz --strip-components=1 -C /usr/local/bin
+      - name: Check all examples
+        run: make check-examples
       - name: Check for unused dependencies
         run: make check-unused-deps
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@cargo-hack
+      - name: Check all feature combinations
+        run: make check-all-features
```
mitinarseny, line 62:
```suggestion
      - name: Check all
        run: make check-all
```

contracts/poa/token/src/contract.rs
```diff
@@ -32,16 +32,16 @@ pub struct Contract {
...
-            icon: Default::default(),
-            reference: Default::default(),
-            reference_hash: Default::default(),
-            decimals: Default::default(),
+            name: String::default(),
+            symbol: String::default(),
+            icon: Option::default(),
+            reference: Option::default(),
+            reference_hash: Option::default(),
+            decimals: u8::default(),
```
mitinarseny:
```suggestion
            name: String::new(),
            symbol: String::new(),
            icon: None,
            reference: None,
            reference_hash: None,
            decimals: 0,
```

mitinarseny (commented):
nits:

.github/workflows/ci.yml
pityjllk: im not sure if i like it, now you know at which step it failed, if its bundled together you need to again view the ci logs... also if we go with single target for check why not include test there ....


also check-all triggers also formatting checks, that requires `taplo` to be installed while its already beeing instaled in `fmt` job. so i would rather keep separate steps (preferred) or bundle everything together into single `check` step ...  



wdyt

.github/workflows/ci.yml
```diff
@@ -52,11 +52,17 @@ jobs:
         run: make check
       - name: Run clippy for defuse
         # Run clippy only on the main contract feature
-        run: cargo clippy --manifest-path ./contracts/defuse/Cargo.toml --features=contract --features=near-sdk/non-contract-usage --no-deps
+        run: cargo clippy --manifest-path ./contracts/defuse/Cargo.toml --features=contract --target wasm32-unknown-unknown
```
mitinarseny: `make check` has a pre-requisite of `make check-contracts`, which already does it, no?

pityjllk: yep its not needed anymore

Makefile
```diff
@@ -3,6 +3,97 @@ DEFUSE_OUT_DIR ?= $(ROOT_DIR)res
...
+	@echo "  clean               Remove build artifacts and cargo clean"
+	@echo "  test                Run all workspace tests"
+	@echo "  check               Run clippy on codebase (codebase + per-contract wasm)"
+	@echo "  check-all           Run all checks"
+	@echo "  fmt                 Format Rust files and Cargo.toml manifests"
+	@echo "  help                Show this help"
+
+.PHONY: clean-out-dir
+clean-out-dir:
+	rm -rf $(DEFUSE_OUT_DIR)
```
mitinarseny, line 24: side note: what if `$(DEFUSE_OUT_DIR)` does not exist?

Makefile
```diff
@@ -36,6 +130,10 @@ $(eval $(shell cargo metadata --format-version=1 | jq -rn \
      ($$vval.container_build_command | join(" ")) as $$non_reproducible_cmd | \
      (if $$reproducible != "" then $$reproducible_cmd else $$non_reproducible_cmd end) as $$cmd | \
      (if $$vkey == "" then "" else ".\($$vkey)" end) as $$suffix | \
+     ($$vval.container_build_command | map(select(startswith("--features="))) | if length > 0 then " " + first else "" end) as $$features_flag | \
+     "$$(eval .PHONY: check-contract/\($$tname))", \
+     "$$(eval check-contracts:: check-contract/\($$tname))", \
+     "$$(eval check-contract/\($$tname):; cargo clippy -p \($$name) --no-deps --target wasm32-unknown-unknown\($$features_flag))", \
```
mitinarseny, line 136: Maybe also add `check-contract/<CONTRACT>/all`?

Makefile
```diff
@@ -47,51 +145,25 @@ $(eval $(shell cargo metadata --format-version=1 | jq -rn \
...
-check:
-	cargo clippy --workspace --all-targets --no-deps
-
-.PHONY: check-fmt
-check-fmt:
-	cargo fmt --all --check
-	RUST_LOG=warn taplo format --check
+$(foreach c,$(CRATES_AT_LEAST_ONE_VARIANT),\
+  $(eval check-all-features-wasm::; \
+    $(CARGO_CHECK_WASM) -p $(call crate_name,$c) --feature-powerset --at-least-one-of $(call crate_features,$c)))
```
mitinarseny, line 167: In general, this PR tries to check every crate we have for WASM target with all supported feature combinations.
This is, of course, a very comprehensive approach. But it seems to me that we're trying to cover all target/feature combinations but then I'm also asking myself: why do we check only host and WASM targets? Why don't we check different host targets as well instead of just the current host target?
Reason for WASM is clear: we use it in contracts. But, if so, then why do we need these additional per-crate checks, if the effectively needed result is already achieved by `check-contracts` make target (I like that!), which is auto-generated from reproducible builds metadata in contracts Cargo.toml?

The reason for me worrying about this is that we will have to maintain these `CRATES_AT_LEAST_ONE_VARIANT`, `CRATES_SKIP_HOST_FEATURES`, `CRATES_HOST_ONLY` and etc in this Makefile, which would only grow in size more and more in the future and it will be a pain to maintain it...

What do you think if we only keep `make check-contracts` and avoid using `cargo hack` entirely?🫠

Makefile
```diff
@@ -3,6 +3,97 @@ DEFUSE_OUT_DIR ?= $(ROOT_DIR)res
...
 
+.PHONY: help
+help:
+	@echo "Usage: make [target] [REPRODUCIBLE=1]"
+	@echo ""
+	@echo "Build targets (use REPRODUCIBLE=1 for reproducible builds):"
+	@$(foreach t,$(ALL_TARGETS),echo "  $(t)";)
+	@echo ""
+	@echo "Other targets:"
+	@echo "  all                 Build all contracts (default)"
```
mitinarseny, line 14: Since we're touching this already and we're going to have outlayer contracts and **services**, maybe it makes sense to introduce `make contracts` target and make it a pre-requizite of `make all`?

Makefile
pityjllk: its fine with `-rf`

Makefile
pityjllk: >In general, this PR tries to check every crate we have for WASM target with all supported feature combinations.

it tries to compiler every crate with every SINGLE feature separately.

TLDR;
if we put that much effort guard impl behind features then it make sense to know that given crate can compile without compiler/linter warnings.

Makefile
pityjllk: we can do that but i would prefer doing it in separate pr, wdyt?

Makefile
pityjllk: added, and listes all check targets as part of make help

Makefile
```diff
@@ -3,6 +3,68 @@ DEFUSE_OUT_DIR ?= $(ROOT_DIR)res
...
+
+.PHONY: check-all
+check-all: check-fmt check-unused-deps check check-examples
+
+.PHONY: fmt
+fmt:
+	cargo fmt --all
+	taplo format
+
+RUSTFLAGS_CHECK = -D warnings
```
mitinarseny: this is not needed anymore, right?

Makefile
```diff
@@ -36,6 +102,12 @@ $(eval $(shell cargo metadata --format-version=1 | jq -rn \
      ($$vval.container_build_command | join(" ")) as $$non_reproducible_cmd | \
      (if $$reproducible != "" then $$reproducible_cmd else $$non_reproducible_cmd end) as $$cmd | \
      (if $$vkey == "" then "" else ".\($$vkey)" end) as $$suffix | \
+     ($$vval.container_build_command | map(select(startswith("--features="))) | if length > 0 then " " + first else "" end) as $$features_flag | \
+     "$$(eval .PHONY: check-contract/\($$tname))", \
+     "$$(eval CHECK_TARGETS += check-contract/\($$tname))", \
+     "$$(eval check-contracts:: check-contract/\($$tname))", \
+     "$$(eval check-contract/\($$tname):; cargo clippy -p \($$name) --no-deps --target wasm32-unknown-unknown\($$features_flag))", \
```
mitinarseny: Let's make it look like scoping of the same command:
```suggestion
     "$$(eval check-contracts/\($$tname):; cargo clippy -p \($$name) --no-deps --target wasm32-unknown-unknown\($$features_flag))", \
```

Makefile
```diff
@@ -36,6 +102,12 @@ $(eval $(shell cargo metadata --format-version=1 | jq -rn \
      ($$vval.container_build_command | join(" ")) as $$non_reproducible_cmd | \
      (if $$reproducible != "" then $$reproducible_cmd else $$non_reproducible_cmd end) as $$cmd | \
      (if $$vkey == "" then "" else ".\($$vkey)" end) as $$suffix | \
+     ($$vval.container_build_command | map(select(startswith("--features="))) | if length > 0 then " " + first else "" end) as $$features_flag | \
+     "$$(eval .PHONY: check-contract/\($$tname))", \
+     "$$(eval CHECK_TARGETS += check-contract/\($$tname))", \
+     "$$(eval check-contracts:: check-contract/\($$tname))", \
```
mitinarseny: Can we make the hierarchy to look like following?
```
check-contracts:: check-contracts/defuse/all
check-contracts/defuse/all:: check-contracts/defuse
check-contracts/defuse/all:: check-contracts/defuse/far
```

Makefile
pityjllk: yes as well as `check-examples` target as `clippy --workspace --all-targets` already covers it

mitinarseny (approved):

PR #248, opened 2026-04-10
feat: outlayer sys

hlgltvnnk:
## Release Notes

* **New Features**
  * Added cryptographic operations support for Ed25519 and secp256k1, enabling public key derivation and message signing functionality.
  * Added WebAssembly runtime and component model tooling support for enhanced platform compatibility.
  * Extended workspace infrastructure with new modular components for improved architecture organization.


crates/outlayer/host/src/ed25519.rs
```diff
@@ -0,0 +1,25 @@
...
+
+// TODO: use defuse crypto?
+pub type Ed25519PublicKey = [u8; 32];
+pub type Ed25519Signature = [u8; 64];
+
+pub trait Ed25519Host {
+    fn ed25519_get_project_public_key() -> Ed25519PublicKey;
+    fn ed25519_derive_public_key(_path: impl AsRef<str>) -> Ed25519PublicKey;
+    fn ed25519_sign(_path: impl AsRef<str>, _msg: impl AsRef<[u8]>) -> Ed25519Signature;
+}
```
mitinarseny:
```suggestion
pub trait Ed25519Host {
    fn ed25519_derive_public_key(&self, path: impl AsRef<str>) -> Ed25519PublicKey;
    fn ed25519_sign(&self, path: impl AsRef<str>, msg: impl AsRef<[u8]>) -> Ed25519Signature;
}
```

crates/outlayer/host/src/secp256k1.rs
```diff
@@ -0,0 +1,25 @@
...
+
+// TODO: use defuse crypto?
+pub type Secp256k1PublicKey = [u8; 64];
+pub type Secp256k1Signature = [u8; 65];
+
+pub trait Secp256k1Host {
+    fn secp256k1_get_project_public_key() -> Secp256k1PublicKey;
+    fn secp256k1_derive_public_key(_path: impl AsRef<str>) -> Secp256k1PublicKey;
+    fn secp256k1_sign(_path: impl AsRef<str>, _msg: impl AsRef<[u8]>) -> Secp256k1Signature;
+}
```
mitinarseny:
```suggestion
pub trait Secp256k1Host {
    fn secp256k1_derive_public_key(&self, path: impl AsRef<str>) -> Secp256k1PublicKey;
    fn secp256k1_sign(&self, path: impl AsRef<str>, msg: impl AsRef<[u8]>) -> Secp256k1Signature;
}
```

crates/outlayer/host/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+#[cfg(feature = "ed25519")]
+pub mod ed25519;
+#[cfg(feature = "secp256k1")]
+pub mod secp256k1;
+
+pub struct DefaultHost;
```
mitinarseny:
```suggestion
#[cfg(not(target_family = "wasm"))]
pub struct WorkerHost {
  ed25519: WorkerEd25519Host,
}
```

crates/outlayer/sdk/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+#[cfg(feature = "ed25519")]
+pub mod ed25519;
+#[cfg(feature = "secp256k1")]
+pub mod secp256k1;
+
+pub struct SysHost;
+
+#[cfg(target_family = "wasm")]
+pub type Host = SysHost;
```
mitinarseny:
```suggestion
#[cfg(target_family = "wasm")]
pub struct SysHost;

#[cfg(target_family = "wasm")]
pub type Host = SysHost;
```

crates/outlayer/sdk/src/lib.rs
mitinarseny, line 1: split by hosts, not by features/worlds
```
./lib.rs
./host/mod.rs
./host/sys.rs
```

crates/outlayer/host/src/crypto/ed25519.rs
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    fn ed25519_derive_public_key(_path: impl AsRef<str>) -> Ed25519PublicKey {
+        unimplemented!("ed25519_derive_public_key is not implemented for DefaultHost");
+    }
+
+    fn ed25519_sign(_path: impl AsRef<str>, _msg: impl AsRef<[u8]>) -> Ed25519Signature {
+        unimplemented!("ed25519_sign is not implemented for DefaultHost");
+    }
+}
```
mitinarseny, line 8:
```suggestion
#[cfg(not(target_family = "wasm"))]
pub struct WorkerEd25519Host;
```

crates/outlayer/sdk/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+[lib]
+crate-type = ["rlib"]
+
+[features]
+ed25519 = ["defuse-outlayer-host/ed25519"]
+secp256k1 = ["defuse-outlayer-host/secp256k1"]
+
+[dependencies]
+defuse-outlayer-host = { workspace = true }
+wit-bindgen.workspace = true
```
mitinarseny:
```suggestion

[target.'cfg(target_family = "wasm")'.dependencies]
wit-bindgen.workspace = true
```

crates/outlayer/wit/crypto.wit
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    record ed25519-signature {
+        bytes: list<u8>,
+    }
+
+    get-project-public-key: func() -> result<ed25519-public-key, string>;
+    derive-public-key: func(path: string) -> result<ed25519-public-key, string>;
+    sign: func(path: string, msg: list<u8>) -> result<ed25519-signature, string>;
+}
```
mitinarseny, line 6:
```suggestion
interface ed25519 {
    derive-public-key: func(path: string) -> list<u8>;
    sign: func(path: string, msg: list<u8>) -> list<u8>;
}
```

crates/outlayer/host/src/lib.rs
hlgltvnnk: Used compile_error instead

Cargo.toml
```diff
@@ -147,6 +158,10 @@ tokio = { version = "1.45", default-features = false }
 tracing = "0.1"
 tracing-subscriber = "0.3"
 
+wasmtime = { version = "36", features = ["component-model"] }
+wasmtime-wasi = "36"
+wit-bindgen = "0.49.0"
```
birchmd: Is there a particular reason to use these versions? Wasmtime has 43 (though 36 is still supported; latest patch was only 4 days ago) and wit-bindgen has 0.55.

crates/outlayer/host/src/ed25519.rs
```diff
@@ -0,0 +1,24 @@
+use crate::WorkerHost;
+use defuse_outlayer_types::ed25519::{Ed25519Host, Ed25519PublicKey, Ed25519Signature};
+
+pub struct WorkerEd25519Host;
+
+impl Ed25519Host for WorkerEd25519Host {
+    fn ed25519_derive_public_key(&self, _path: impl AsRef<str>) -> Ed25519PublicKey {
+        unimplemented!("ed25519_derive_public_key is not implemented for WorkerHost");
```
birchmd, line 8: I assume this is a placeholder and the impl will be included eventually.

crates/outlayer/types/src/ed25519.rs
```diff
@@ -0,0 +1,8 @@
+// TODO: use defuse crypto?
+pub type Ed25519PublicKey = [u8; 32];
+pub type Ed25519Signature = [u8; 64];
+
+pub trait Ed25519Host {
```
birchmd, line 5: I don't love duplicating the trait definition here as well as in `wit`. I think it could be harder to maintain because all changes need to be duplicated in both places as well.

It feels like `wit` is supposed to be the single source of truth and the `generate!` macro will automatically create a trait definition if the interface is exported or function definitions if the interface is imported. Based on this I think we want to define two worlds for each curve: one where the interface is exported and one where it is imported. The host would invoke `generate!` on the world with the export and provide the implementation while the sdk would use `generate!` on the world with the import as is currently the case.

All that said, these trait definitions are simple and  we don't expect the interface to change very often after the initial version so maybe it is not worth the trouble to try to have a single source of truth based on `wit`. I'm just raising it as a potential consideration.

mitinarseny: Hey @birchmd, thanks for the review!

The initial idea behind having these traits is to make WASMs depend only on them and `pub static Host` that implements them instead of calling host-functions directly. 

With that, unit-testing would look cleaner and we can set that `Host` to whatever we want, instead of mocking `extern "C"` functions that reconstructs arguments back to their original types and operates on them (like it's done in near-sdk).

Do you see any downsides of this approach?

Cargo.toml
mitinarseny: This is due to MSRV being 1.86 for this repo. As of now, we decided to go with the old version. Once nearcore 2.12 is on mainnet, we'll be able to use newer rust version, and thus, a newer wasmtime.

crates/outlayer/types/src/ed25519.rs
birchmd: Yes, I agree it is better to have trait definitions separate from `extern` functions for all the reasons you said. I just thought that was achievable with definitions coming from `wit` by having one world export and a separate world import. But if that's not how it works or if we don't want to clutter the `wit` file with extra worlds then that's also fine. I don't have too strong feelings about it. Avoiding the duplicate definition in Rust and `wit` is the only Southfield argument for trying to have a single source of truth.

crates/outlayer/types/src/lib.rs
mitinarseny, line 1: Let's rename `types` crate to `host`, and `host` to `worker-host`.

crates/outlayer/host/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+#[cfg(target_family = "wasm")]
+compile_error!(
+    r#"The `host` crate is only meant to be used in a non-WebAssembly environment.
+    If you want to use host functionality in a WebAssembly environment,
+    please use the `defuse_outlayer_sdk` crate instead.
+"#
+);
```
mitinarseny: Let's remove this compilation error, it's okay to compile everything to WASM except for things that cannot be compiled to that target

crates/outlayer/sdk/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+mod host;
+
+pub use defuse_outlayer_types as types;
+
+#[cfg(not(target_family = "wasm"))]
+pub use defuse_outlayer_host as workers;
```
mitinarseny:
```suggestion
```

crates/outlayer/sdk/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+mod host;
+
+pub use defuse_outlayer_types as types;
```
mitinarseny:
```suggestion
pub use defuse_outlayer_host as host;
```

crates/outlayer/sdk/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+
+#[cfg(not(target_family = "wasm"))]
+pub use defuse_outlayer_host as workers;
+
+#[cfg(target_family = "wasm")]
+pub type Host = host::sys::SysHost;
+
+#[cfg(not(target_family = "wasm"))]
+pub type Host = defuse_outlayer_host::WorkerHost;
```
mitinarseny: It seems that it's better to make `Host` a static variable. Otherwise, since `SysHost` doesn't need any data, you won't be able to `WorkerHost` that will be propagated to all invocations in unit tests

crates/outlayer/sys/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+#[cfg(not(any(feature = "ed25519", feature = "secp256k1")))]
+compile_error!(
+    r#"At least one of these features should be enabled:
+- "ed25519"
+- "secp256k1"
+"#
+);
```
mitinarseny: remove this compilation error pls

Makefile
```diff
@@ -55,7 +54,7 @@ help:
 	@$(foreach t,$(ALL_TARGETS),echo "  $(t)";)
 	@echo ""
 	@echo "Other targets:"
-	@echo "  all              Build all contracts (default)"
+	@echo "  all              Build all contracts and outlayer (default)"
```
mitinarseny: We don't have outlayer binary and this Makefile doesn't build it yet. Revert this change pls

crates/outlayer/host/src/ed25519.rs
hlgltvnnk: Hi, thanks for the review
Yes, the worker host implementation will be in the next pr and will include functionality that will be reused in the vm runner and in the guest wasm unit tests

crates/outlayer/types/src/lib.rs
hlgltvnnk: I would like to have not only methods for the host here - but in general types that can be reused throughout the outlayer project, including contracts and the balancer

mitinarseny:
@claude review

crates/outlayer/host/src/lib.rs
hlgltvnnk: This crate is designed for interfaces that will be used by the host/tests and should not be used as part of wasm (and it is not expected to add functionality that could be reused by wasm) - we explicitly prohibit its use, do you think it would be better to allow usage of this crate for wasm guest? what are the use cases?

crates/outlayer/host/src/crypto/secp256k1.rs
```diff
@@ -0,0 +1,8 @@
+// TODO: use defuse crypto?
+pub type Secp256k1PublicKey = [u8; 64];
```
gitmp01, line 2: Just a note: we would need also the chain code for the offline address generation

hlgltvnnk: We don't know yet whether we will use hierarchical derivation (as currently implemented in the pds) - perhaps we will use flat derivation (since  standart bip32 cant be easily applied to ed25519) which doesn't require chaincode

crates/outlayer/host/src/crypto/secp256k1.rs
```diff
@@ -0,0 +1,8 @@
...
+pub type Secp256k1Signature = [u8; 65];
+
+pub trait Secp256k1Host {
+    fn secp256k1_derive_public_key(&self, path: impl AsRef<str>) -> Secp256k1PublicKey;
```
gitmp01, line 6: Sorry if I jump here again, but wouldn't usage become simpler if we define a common interface like 

```rs
pub trait Common {
  fn derive_public_key(&self, path: impl AsRef<str>) -> Vec<u8>
  fn sign(&self, path: impl AsRef<str>, msg: impl AsRef<[u8]>) -> Vec<u8>;
}
```

And let `Ed25519Host` & `Secp256k1Host` implement that?

This way separation on the client happens during object instantiation, like

```rs
#[cfg(feature = "evm")]
let host = Secp256k1Host {}

#[cfg(feature = "sol")]
let host = Ed25519Host {}

// call host.derive_public_key()
```

instead of having something like:

```rs
let pub_key = match self {
    #[cfg(feature = "evm")]
    ChainType::Evm => x.secp256k1_derive_public_key(EVM_HD_PATH),
    #[cfg(feature = "sol")]
    ChainType::Solana => x.ed25519_derive_public_key("")
}
```

mitinarseny (dismissed):

crates/outlayer/host/src/crypto/secp256k1.rs
hlgltvnnk: With Tee worker we want to enable guest wasm to call the derivation and signature host functions for any curve (without splitting into features  for particular curves as this is basic functionality) - by calling HOST.some_curve_sign()
And already at the guest wasm level you can introduce splitting by features

mitinarseny (dismissed):

crates/outlayer/host/src/crypto/secp256k1.rs
gitmp01: I see, anyway my concern was more on usability on the guest side, but it isn't that important

mitinarseny (approved):

PR #252, opened 2026-04-20
refactor: sdk host functions

mitinarseny:
## Summary by CodeRabbit

* **New Features**
  * Added public derive_public_key() and sign() functions for Ed25519.
  * Added public derive_public_key() and sign() functions for Secp256k1.

* **Refactor**
  * Restructured host/crypto surface and SDK host layout for clearer separation of crypto and mock paths.
  * Simplified worker host surface and replaced prior system-backed crypto surface with a mock-backed approach.


crates/outlayer/sdk/src/host/crypto/secp256k1.rs
```diff
@@ -0,0 +1,21 @@
...
+        .try_into()
+        .expect("secp256k1 public key must be 64 bytes");
+
+    #[cfg(not(target_family = "wasm"))]
+    return crate::host::mock::HOST.with_borrow(|h| h.secp256k1_derive_public_key(path));
+}
+
+pub fn sign(path: impl AsRef<str>, msg: impl AsRef<[u8]>) -> Secp256k1Signature {
+    #[cfg(target_family = "wasm")]
+    return ::defuse_outlayer_sys::crypto::ed25519::sign(path.as_ref(), msg.as_ref())
```
hlgltvnnk: secp256k1

crates/outlayer/host/src/lib.rs
```diff
@@ -1 +1,6 @@
 pub mod crypto;
+
```
hlgltvnnk, line 2: This will be used only in worker-host - may be move it there?

hlgltvnnk (changes requested):

crates/outlayer/host/src/lib.rs
mitinarseny: I was thinking the same, but then I realized that we have these `Ed25519PublicKey` type aliases and might have other structs defined in the future, which should be returned by host-function-wrappers in outlayer-sdk as well

PR #253, opened 2026-04-20
feat(outlayer): vm runner

hlgltvnnk:
## Release Notes

**New Features**
* Added VM runtime for executing WebAssembly components with configurable fuel and memory limits.
* Introduced state management module with support for ed25519 and secp256k1 cryptographic operations.
* Added WASI P2 backend support for WebAssembly component execution.
* New error types for handling execution failures and validation errors.


crates/outlayer/state/src/crypto/secp256k1.rs
```diff
@@ -0,0 +1,20 @@
+use defuse_outlayer_host::crypto::secp256k1::Secp256k1Host;
+use defuse_outlayer_sys::host::outlayer;
+
+use crate::HostState;
+
+impl outlayer::crypto::secp256k1::Host for HostState {
```
pityjllk, line 6: `Host` is kind of misleading for me but it feels like it s more like cryto related trait than `Host`

crates/outlayer/sys/src/lib.rs
```diff
@@ -1,7 +1,30 @@
...
+#[cfg(feature = "host")]
+pub mod host {
+    wasmtime::component::bindgen!({
+        path: "../wit",
+        world: "imports",
+    });
+
+    pub use self::outlayer::*;
+
+    pub trait Host:
```
pityjllk: Maybe more like HostFunctions, or Host interface, not a big fan of this name tbh 🙃

crates/outlayer/vm-runner/src/backend/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+
+    fn setup_linker<H: Host>(linker: &mut Linker<HostCtx<Self::State, H>>) -> Result<()>;
+
+    fn build_state(
+        stdin: MemoryInputPipe,
+        stdout: MemoryOutputPipe,
+        stderr: MemoryOutputPipe,
+    ) -> Self::State;
+
+    async fn call_run<H: Host>(
```
pityjllk: maybe just `run` ?

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,199 @@
...
+fn classify_result(
+    program_result: anyhow::Result<()>,
+    stderr: &MemoryOutputPipe,
+) -> Result<(), ExecutionError> {
+    let stderr = String::from_utf8_lossy(&stderr.contents()).into_owned();
+
+    let trap = match program_result {
+        Ok(()) => return Ok(()),
+        Err(e) => e,
+    };
```
pityjllk:
```suggestion

    let trap = match program_result {
        Ok(()) => return Ok(()),
        Err(e) => e,
    };
    
        let stderr = String::from_utf8_lossy(&stderr.contents()).into_owned();
```
to avoid allocation when not needed

pityjllk (commented):
few minor comments

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,199 @@
...
+        &self,
+        component: &Component,
+        host_state: H,
+        input: I,
+    ) -> Result<ExecutionOutcome, VmError>
+    where
+        H: Host + 'static,
+        I: serde::Serialize + Send,
+    {
+        let linker = self.create_linker::<H>()?;
```
pityjllk: maybe we could reuse the linker (pass it as parameter here as well instead of creating it each time?)

crates/outlayer/state/src/crypto/secp256k1.rs
hlgltvnnk: it is autogenerated trait, but we can create use alias over it

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+        config.guard_before_linear_memory(true);
+        config.memory_guard_size(MEMORY_GUARD_SIZE);
+        // NOTE: this is enabling async in host-defined functions
+        config.async_support(true);
+        config.consume_fuel(true);
+
+        Self {
+            config,
+            fuel_limit: None,
+            memory_limit: None,
```
mitinarseny, line 61: Use defaults instead of `Option`s

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        Ok(VmRuntime {
+            fuel_limit: self.fuel_limit.unwrap_or(DEFAULT_FUEL_LIMIT),
+            memory_limit,
+            linker,
+        })
+    }
+}
+
+fn create_linker<H: HostFunctions + 'static>(engine: &Engine) -> Result<Linker<HostCtx<H>>> {
```
mitinarseny, line 108: make it a private method of `VmRuntimeBuilder` or even use it inline in the `build()` method above

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+        store: &mut Store<HostCtx<H>>,
+        component: &Component,
+    ) -> anyhow::Result<()> {
+        let command = Command::instantiate_async(&mut *store, component, &self.linker).await?;
+
+        command
+            .wasi_cli_run()
+            .call_run(&mut *store)
+            .await?
+            .map_err(|()| anyhow::anyhow!("component run() returned Err(())"))
```
mitinarseny, line 216: 1. this will not be counted as a trap, right?
2. why not handling stderr here?

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+fn classify_result(
+    program_result: anyhow::Result<()>,
+    stderr: &MemoryOutputPipe,
+) -> Result<(), ExecutionError> {
+    let trap = match program_result {
+        Ok(()) => return Ok(()),
+        Err(e) => e,
+    };
+
+    let stderr = String::from_utf8_lossy(&stderr.contents()).into_owned();
```
mitinarseny, line 251: why does stderr has to be a valid string?

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+/// The host environment is defined by the `HostFunctions` trait,
+/// which must be implemented by the user and provided as a type
+/// parameter to the builder.
+pub struct VmRuntime<H: HostFunctions + 'static> {
+    fuel_limit: u64,
+    memory_limit: usize,
+    linker: Linker<HostCtx<H>>,
+}
+
+impl<H: HostFunctions + 'static> VmRuntime<H> {
```
mitinarseny, line 140: add `.builder()` and `.new()` methods directly to `VmRuntime`

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// ```
+    #[instrument(skip_all)]
+    pub async fn execute(
+        &self,
+        component: &Component,
+        host_state: H,
+        input: impl AsRef<[u8]>,
+    ) -> Result<ExecutionOutcome, ExecutionError> {
+        let stdin = MemoryInputPipe::new(input.as_ref().to_vec());
+        let stdout = MemoryOutputPipe::new(STDOUT_MAX_SIZE);
```
mitinarseny, line 175: can we avoid buffering the entire input, so that it can be propagated to the `execute` method? E.g. what if the input is a large file and WASM is supposed to process it line-by-line. Can we avoid storing it entirely in-memory, as well

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+
+        command
+            .wasi_cli_run()
+            .call_run(&mut *store)
+            .await?
+            .map_err(|()| anyhow::anyhow!("component run() returned Err(())"))
+    }
+}
+
+fn process_execution_result(
```
mitinarseny, line 220: why don't we return stdout on failure as well?

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub async fn execute(
+        &self,
+        component: &Component,
+        host_state: H,
+        input: impl AsRef<[u8]>,
+    ) -> Result<ExecutionOutcome, ExecutionError> {
+        let stdin = MemoryInputPipe::new(input.as_ref().to_vec());
+        let stdout = MemoryOutputPipe::new(STDOUT_MAX_SIZE);
+        let stderr = MemoryOutputPipe::new(STDERR_MAX_SIZE);
+        let wasi_ctx = WasiCtx::builder()
```
mitinarseny, line 177: The same about buffering applies to stdout and stderr: can we allow callers to pass an object of a type that implements some `Write` trait, so that it decides what to do further with the stdout and stderr being written to by WASM.

crates/outlayer/vm-runner/src/outcome.rs
```diff
@@ -0,0 +1,7 @@
+/// Outcome of executing a component in the VM runtime
+#[derive(Debug)]
+pub struct ExecutionOutcome {
+    pub output: Vec<u8>,
+    pub stderr: Vec<u8>,
```
mitinarseny, line 5: the naming is inconsistent: `output` vs `stderr`
let's make it `stdout` and `stderr`

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+    ///  async fn example() -> anyhow::Result<()> {
+    ///     let wasm_binary = std::fs::read("component.wasm")?;
+    ///     let runner = VmRuntimeBuilder::<HostState>::new().build()?;
+    ///     let component = runner.compile(&wasm_binary)?;
+    ///     let outcome = runner.execute(&component, HostState::default(), b"input").await?;
+    ///    Ok(())
+    /// }
+    /// ```
+    #[instrument(skip_all)]
+    pub async fn execute(
```
mitinarseny, line 168: expose another `run` method that will be a wrapper: `compile` + `execute`?

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+            .memory_reservation(memory_limit.try_into().unwrap());
+
+        let engine = Engine::new(&self.config)?;
+        let linker = create_linker::<H>(&engine)?;
+
+        Ok(VmRuntime {
+            fuel_limit: self.fuel_limit.unwrap_or(DEFAULT_FUEL_LIMIT),
+            memory_limit,
+            linker,
+        })
```
mitinarseny, line 104: If both `fuel_limit` and `memory_limit` are used in `VmRunner` only in runtime but not at initialization time, then why do you need a separate builder struct here?
```rust
impl VmRunner {
    pub fn new(config: Config) -> Self {
        // ...
    }

    pub fn with_fuel_limit(mut self, limit: u64) -> Self {
        // ...
    }

    pub fn with_memory_limit(mut self, limit: usize) -> Self {
        // ...
    }
}
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// Builds the `VmRuntime` with the specified configuration
+    /// If fuel limit or memory limit are not set, default values will be used
+    pub fn build(mut self) -> Result<VmRuntime<H>> {
+        let memory_limit = self.memory_limit.unwrap_or(DEFAULT_MEMORY_LIMIT);
+        // NOTE: set initial chunk of virtual memory that a linear memory
+        // may grow into limited to allow multiple linear memories to be
+        // instantiated without exhausting host resources
+        self.config
+            .memory_reservation(memory_limit.try_into().unwrap());
+
```
mitinarseny, line 96: Let's play with all config optimization/limiting params later, not in this PR.

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: That is not trap - but component error, as doc says it should be returned if wasm component finished with Error es result of its execution, but in practice i can only catch the outer error (it is converted to I32Exit) -  so I havent been able to catch it yet. Stderr handled after when errors  are downcasted

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: i will just comment that

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: moved  it to caller Context

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: moved  it to caller Context (not vm runner responsibility)

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: moved  it to caller Context

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: But what if more parameters for initialization appear - I like the idea of ​​separating responsibility for initialization and for the execution logic itself into different entities, it is more concise

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,368 @@
...
+    if let Some(exit) = trap.downcast_ref::<wasmtime_wasi::I32Exit>() {
+        // exit(0) is equivalent to a clean return from main
+        return match exit.0 {
+            0 => Ok(()),
+            code => Err(VmError::NonZeroExit(code)),
+        };
+    }
+
+    let trap_code = trap.chain().find_map(|e| e.downcast_ref::<Trap>().copied());
```
mitinarseny: maybe no `.chain().find_map()`?

crates/outlayer/vm-runner/src/error.rs
```diff
@@ -0,0 +1,27 @@
+use wasmtime::Trap;
+
+/// Errors that can occur during execution of a WASI component
+#[derive(thiserror::Error, Debug)]
+pub enum VmError {
+    /// An error that occurred during initialization of the runtime
+    #[error("initialization error: {0}")]
+    Init(#[source] anyhow::Error),
```
mitinarseny: remove

hlgltvnnk: updated - but I dont like that we mix our error types (compile/execute) with anyhow (builder functions) in this case

crates/outlayer/vm-runner/src/context.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl WasiView for WasiP2State {
+    fn ctx(&mut self) -> WasiCtxView<'_> {
+        WasiCtxView {
+            ctx: &mut self.ctx,
+            table: &mut self.table,
+        }
+    }
+}
```
pityjllk, line 71: seems to be used only in this file. maybe consider just adding ctx into HostCtx and get rid of WasiP2State?

crates/outlayer/vm-runner/src/outcome.rs
```diff
@@ -0,0 +1,8 @@
+use crate::error::ExecutionError;
+
+/// Outcome of executing a component in the VM runtime
+#[derive(Debug)]
+pub struct ExecutionOutcome {
+    pub fuel_consumed: u64,
+    pub guest_error: Option<ExecutionError>,
```
pityjllk: aftert giving it another look, maybe lets simply have `execution_result: Result<(), ExecutionError>`, wdyt?

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+    /// Async support and fuel metering are always enabled and cannot be
+    /// disabled. Use [`fuel_limit`](Self::fuel_limit) and
+    /// [`memory_limit`](Self::memory_limit) to override the defaults
+    pub fn new() -> Self {
+        let mut config = Config::new();
+        config.guard_before_linear_memory(true);
+        config.memory_guard_size(MEMORY_GUARD_SIZE);
+        // NOTE: this is enabling async in host-defined functions
+        config.async_support(true);
+        config.consume_fuel(true);
```
pityjllk: i think it makes sense to introduce "our" config type - like with these params that you explicitally set here and provide builder pattern. My reasoning is that we will rather want to do sth like
```   
fn main() {
let config = clap::parse()
let vm_runnner::new(config.runer_config);
let tee_service::new(vm_runner)
}
```

rather than using builder for each paramter

pityjllk (changes requested):

crates/outlayer/vm-runner/src/context.rs
hlgltvnnk: I think dividing into different contexts looks more concise and readable

crates/outlayer/vm-runner/src/outcome.rs
hlgltvnnk: I don't quite understand the point of keeping result if a successful result is never processed

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: All of these parameters are optional, there are only 2 of them at the moment, but I think that as the number of these parameters increases, it would be more convenient to build this from a builder perspective than a config, but may be im wrong, wdyt?

crates/outlayer/cli/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "defuse-outlayer-cli"
```
mitinarseny: Let's rename the package and move it to `crates/outlayer/vm-runner/cli`: we will have more CLIs, not only this one

```suggestion
name = "defuse-outlayer-vm-runner-cli"
```

crates/outlayer/cli/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+    about = "Execute a WASI component with a custom host environment",
+    long_about = "\n
+    Runs a WASIP2 component, wiring its stdin/stdout/stderr to the host \n
+    Input can be piped from stdin or supplied via --input-file.\n
+    Execution is bounded by configurable fuel and memory limits.
+    "
+)]
+struct Args {
+    /// Application ID to use in the host context
+    app_id: AppId<'static>,
```
mitinarseny: Let's make it optional argument and derive it by default from OutlayerApp state-init with`000..000` admin-id, WASM hash and `data:application/wasm;base64,<WASM_BASE64>`

crates/outlayer/cli/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// Path to the WebAssembly component to execute
+    path: PathBuf,
+
+    /// Maximum number of WebAssembly instructions the component may execute
+    #[clap(long, short)]
+    fuel_limit: Option<u64>,
```
mitinarseny:
```suggestion
    /// Maximum number of WebAssembly instructions the component may execute
    #[clap(long, short)]
    fuel: Option<u64>,
```

crates/outlayer/cli/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// Maximum memory the component may use, in bytes
+    #[clap(long, short)]
+    memory_limit: Option<usize>,
```
mitinarseny: Lets add support for human-readable byte-count representations via [bytesize](https://crates.io/crates/bytesize) crate:
```suggestion
    /// Maximum memory the component may use
    #[clap(long, short)]
    memory: Option<ByteSize>,
```

crates/outlayer/cli/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let args = Args::parse();
+
+    let seed = env::var("OUTLAYER_SEED")
+        .map_err(|_| anyhow::anyhow!("OUTLAYER_SEED env var is not set"))?;
```
mitinarseny: Why seed is configured via environment variables and other - via CLI arguments? Let's unify it and use CLI args only

crates/outlayer/cli/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    let state = State::new(
+        HostContext {
+            app_id: args.app_id,
+        },
+        Cow::Owned(InMemorySigner::from_seed(seed.as_bytes())),
```
mitinarseny: Let's make seed to be hex-encoded vector of bytes, not necessarily a UTF-8 string

crates/outlayer/cli/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+    let mut builder = VmRuntime::<State>::builder();
+    if let Some(fuel_limit) = args.fuel_limit {
+        builder = builder.fuel_limit(fuel_limit);
+    }
+    if let Some(memory_limit) = args.memory_limit {
+        builder = builder.memory_limit(memory_limit);
+    }
+    let runner = builder.build()?;
+
+    let wasm_binary = std::fs::read(args.path)?;
```
mitinarseny:
```suggestion
    let wasm_binary = std::fs::read(args.path).context("failed to read {}", args.path)?;
```

crates/outlayer/cli/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+    let component = runner.compile(&wasm_binary)?;
```
mitinarseny:
```suggestion
    let component = runner.compile(&wasm_binary).context("compile")?;
```

crates/outlayer/cli/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+    println!("-------------------------------------------");
+    println!(
+        "Execution completed. Fuel consumed: {}",
+        outcome.fuel_consumed
+    );
+    if let Some(error) = outcome.guest_error {
+        eprintln!("Failed with error: {error:?}");
+    } else {
+        println!("Succeeded without errors");
+    }
```
mitinarseny: 1. Don't write to stdout anything but WASM's output.
2. Write to stderr only if error occurred while running WASM.
3. Do not report fuel consumption here. We don't charge for fuel right now. In the future, we will might need a structured output with execution details, not just plain println's.

crates/outlayer/cli/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[lints]
+workspace = true
+
+[dependencies]
+anyhow.workspace = true
+clap = { workspace = true, features = ["derive"] }
+
+defuse-outlayer-host = { workspace = true }
+defuse-outlayer-vm-runner = { workspace = true }
```
mitinarseny: It seems to me that `vm-runner` should re-export `pub use defuse_outlayer_host as host`

crates/outlayer/sdk/src/host/crypto/ed25519.rs
```diff
@@ -22,6 +22,8 @@ pub fn derive_public_key(path: impl AsRef<str>) -> PublicKey {
     #[cfg(target_family = "wasm")]
     {
         sys::crypto::ed25519::derive_public_key(path.as_ref())
+            .try_into()
+            .expect("invalid length")
```
mitinarseny: it is already done below

```suggestion
```

crates/outlayer/vm-runner/src/error.rs
```diff
@@ -0,0 +1,21 @@
+use wasmtime::Trap;
+
+/// Errors that can occur during execution of a WASI component
+#[derive(thiserror::Error, Debug)]
+pub enum ExecutionError {
+    /// The component called `process::exit` with exit code
+    #[error("component exited with code {0}")]
```
mitinarseny:
```suggestion
    #[error("component exited with code: {0}")]
```

crates/outlayer/vm-runner/src/error.rs
```diff
@@ -0,0 +1,21 @@
...
+    #[error("wasm trap: {0}")]
+    Trap(Trap),
+
+    /// The component execution failed
+    #[error("component execution failed")]
+    Failed,
+
+    /// An error that does not map to a known trap code
+    #[error("component trapped: {0}")]
+    Unknown(#[source] anyhow::Error),
```
mitinarseny: `Failed` is already obvious since there was an error. It can be returned as `ExecutionError::Custom(anyhow!("failed"))`

```suggestion
    /// An error that does not map to a known trap code
    #[error(transparent)]
    Custom(#[source] anyhow::Error),
```

crates/outlayer/vm-runner/src/error.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Errors that can occur during execution of a WASI component
+#[derive(thiserror::Error, Debug)]
+pub enum ExecutionError {
+    /// The component called `process::exit` with exit code
+    #[error("component exited with code {0}")]
+    NonZeroExit(i32),
+
+    /// The component raised a WebAssembly trap
+    #[error("wasm trap: {0}")]
```
mitinarseny: This error already denotes WASM error
```suggestion
    #[error("trap: {0}")]
```

crates/outlayer/cli/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+    );
+
+    let mut builder = VmRuntime::<State>::builder();
+    if let Some(fuel_limit) = args.fuel_limit {
+        builder = builder.fuel_limit(fuel_limit);
+    }
+    if let Some(memory_limit) = args.memory_limit {
+        builder = builder.memory_limit(memory_limit);
+    }
+    let runner = builder.build()?;
```
mitinarseny:
```suggestion
    let runner = builder.build().context("runtime: build")?;
```

crates/outlayer/host/src/lib.rs
```diff
@@ -22,3 +22,15 @@ impl<'a> State<'a> {
...
+    bindings::outlayer::crypto::ed25519::Host + bindings::outlayer::crypto::secp256k1::Host + Send
+{
+}
+
+impl<T> HostFunctions for T where
+    T: bindings::outlayer::crypto::ed25519::Host
+        + bindings::outlayer::crypto::secp256k1::Host
+        + Send
+{
+}
```
mitinarseny: Why is it needed in this trait? It seems that it is `vm-runner` who requires a type to implement these traits. If so, then put it in the crate that needs it

Cargo.toml
```diff
@@ -50,6 +50,8 @@ members = [
   "crates/outlayer/primitives",
   "crates/outlayer/sys",
   "crates/outlayer/sdk",
+  "crates/outlayer/vm-runner",
+  "crates/outlayer/cli",
```
mitinarseny: pls sort

crates/outlayer/vm-runner/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+[dependencies]
+anyhow.workspace = true
+thiserror.workspace = true
+
+wasmtime = { workspace = true, features = ["cranelift"] }
+wasmtime-wasi.workspace = true
+
+tracing = { workspace = true }
+
+defuse-outlayer-host = { workspace = true }
```
mitinarseny: Style nits:
2. Put deps from our workspace first
3. Use `.workspace = true` for consistency

crates/outlayer/vm-runner/src/error.rs
```diff
@@ -0,0 +1,21 @@
...
+    Trap(Trap),
```
mitinarseny:
```suggestion
    Trap(#[from] Trap),
```

crates/outlayer/vm-runner/src/outcome.rs
```diff
@@ -0,0 +1,8 @@
+use crate::error::ExecutionError;
+
+/// Outcome of executing a component in the VM runtime
+#[derive(Debug)]
```
mitinarseny, line 11:
```suggestion
#[derive(Debug, Clone, Eq, PartialEq, Hash)]
```

crates/outlayer/vm-runner/src/outcome.rs
mitinarseny: Then maybe add a method:
```rust
impl ExecutionOutcome {
    pub fn into_result(self) -> Result<(), ExecutionError> {
        if let Some(err) = self.guest_error {
            return Err(err);
        }
        Ok(())
    }
}
```

mitinarseny: Naming "gues_error" suggests that it comes from the guest, but it's not really, isn't it? For example, the trap can come from host implementation of a host-function, which comes from the host... So maybe rename it to just `error`?

```suggestion
    pub error: Option<ExecutionError>,
```

crates/outlayer/vm-runner/src/outcome.rs
```diff
@@ -0,0 +1,8 @@
...
+pub struct ExecutionOutcome {
```
mitinarseny, line 12:
```suggestion
#[non_exhaustive]
pub struct ExecutionOutcome {
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+        }
+    }
+}
+
+/// A builder for configuring and creating a `VmRuntime`
+/// with a custom host environment
+pub struct VmRuntimeBuilder<H: HostFunctions + 'static> {
+    config: Config,
+    fuel_limit: u64,
+    memory_limit: usize,
```
mitinarseny: I'm sorry for my [previous comment](https://github.com/near/intents/pull/253#discussion_r3118296529), but let's make limits optional here and apply them only if specified. The reason is that in CLI they're optional, too. And not specifying them in CLI makes the user think that "there are no limits". And it sounds as a sane default, especially for CLI usage

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+
+    /// Sets the maximum physical memory the components linear memory may use
+    ///
+    /// Attempts to grow beyond this limit will trap. Defaults to
+    /// [`DEFAULT_MEMORY_LIMIT`] if not set
+    #[must_use]
+    pub const fn memory_limit(mut self, memory_limit: usize) -> Self {
+        self.memory_limit = memory_limit;
+        self
+    }
```
mitinarseny, line 74: Let's move fuel- and memory-related limits into `Context`. Reasoning:
`VmRuntime` is supposed to be built only once for TEE worker's lifetime. For each request we will execute WASMs separately. For each such request we might want to set custom limits, based on higher-level logic

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+    fuel_limit: u64,
+    memory_limit: usize,
+    linker: Linker<HostCtx<H>>,
+}
+
+impl<H: HostFunctions + 'static> VmRuntime<H> {
+    /// Creates a new `VmRuntime` with the default configuration
+    pub fn new() -> anyhow::Result<Self> {
+        Self::builder().build()
+    }
```
mitinarseny: Implement `Default`, too

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+    /// ```rust,no_run
+    /// use defuse_outlayer_host::State;
+    /// use defuse_outlayer_vm_runner::{Context, VmRuntimeBuilder};
+    /// use wasmtime_wasi::p2::pipe::{MemoryInputPipe, MemoryOutputPipe};
+    ///
+    /// # async fn example() -> anyhow::Result<()> {
+    ///     let state = State::new(
+    ///         HostContext { app_id },
+    ///         Cow::Owned(InMemorySigner::from_seed(seed)),
+    ///     );
```
mitinarseny:
```suggestion
    /// # async fn main() -> anyhow::Result<()> {
    /// let state = State::new(
    ///     HostContext { app_id },
    ///     Cow::Owned(InMemorySigner::from_seed(seed)),
    /// );
```

crates/outlayer/vm-runner/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+mod error;
+mod host;
+mod outcome;
+mod runtime;
+
+pub use error::ExecutionError;
+pub use outcome::ExecutionOutcome;
+pub use runtime::{Context, DEFAULT_FUEL_LIMIT, DEFAULT_MEMORY_LIMIT, VmRuntime, VmRuntimeBuilder};
```
mitinarseny:
```suggestion
pub use self::{error::*, host::*, outcome::*, runtime::*};

pub use wasmtime;
pub use wasmtime_wasi;
```

crates/outlayer/vm-runner/src/context.rs
```diff
@@ -0,0 +1,61 @@
+use defuse_outlayer_host::HostFunctions;
```
mitinarseny:
```suggestion
pub use default_outlayer_host as host;
use defuse_outlayer_host::HostFunctions;
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+use std::marker::PhantomData;
+use tracing::instrument;
+use wasmtime::component::{Component, HasSelf, Linker};
+use wasmtime::{Config, Engine, Store, StoreLimitsBuilder, Trap};
+use wasmtime_wasi::cli::{StdinStream, StdoutStream};
+use wasmtime_wasi::{WasiCtx, p2::bindings::Command};
+
+use crate::error::ExecutionError;
+use crate::host::HostCtx;
+use crate::outcome::ExecutionOutcome;
```
mitinarseny: merge imports pls

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+    /// use wasmtime_wasi::p2::pipe::{MemoryInputPipe, MemoryOutputPipe};
+    ///
+    /// # async fn example() -> anyhow::Result<()> {
+    ///     let state = State::new(
+    ///         HostContext { app_id },
+    ///         Cow::Owned(InMemorySigner::from_seed(seed)),
+    ///     );
+    ///
+    /// let stdout = MemoryOutputPipe::new(4 * 1024 * 1024);
+    /// let stderr = MemoryOutputPipe::new(64 * 1024);
```
mitinarseny:
```suggestion
    /// let stdout = MemoryOutputPipe::new(4 * 1024 * 1024); // 4 MB
    /// let stderr = MemoryOutputPipe::new(64 * 1024); // 64 KB
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+    ///     stderr.clone(),
+    ///     state,
+    /// );
+    ///
+    /// let runner = VmRuntimeBuilder::<State>::new().build()?;
+    /// let wasm = std::fs::read("component.wasm")?;
+    /// let component = runner.compile(&wasm)?;
+    /// runner.execute(ctx, &component).await?;
+    ///
+    /// println!("{}", String::from_utf8_lossy(&stdout.contents()));
```
mitinarseny:
```suggestion
    /// let stdout = stdout.contents();
    /// let stderr = stderr.contents();
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+    ///     Wrapper { inner: shared.clone() },
+    /// );
+    ///
+    /// let runner = VmRuntimeBuilder::<Wrapper>::new().build()?;
+    /// let wasm = std::fs::read("component.wasm")?;
+    /// let component = runner.compile(&wasm)?;
+    /// runner.execute(ctx, &component).await?;
+    ///
+    /// println!("{}", String::from_utf8_lossy(&stdout.contents()));
+    /// # Ok(()) }
```
mitinarseny: This example is misleading, since our host should never be concurrent and/or share resources with other host initialized for other component(s)

crates/outlayer/vm-runner/src/context.rs
```diff
@@ -0,0 +1,61 @@
...
+/// The host context passed to the component, containing both
+/// the WASI state and the custom host state
+///
+/// Used as the context for the linker when instantiating the component, and
+/// passed to host functions when called by the component
+pub struct HostCtx<T: HostFunctions> {
+    wasi_state: WasiP2State,
+    host_state: T,
+    limits: StoreLimits,
+}
```
mitinarseny, line 24: I suggest not to expose low-level `Host` traits for raw bindings. Instead, we should follow the approach described [here](https://docs.rs/wasmtime/latest/wasmtime/component/trait.HasData.html#example-host-for-u-mylibrary-using-storet) in `defuse-outlayer-host` crate:
```rust
pub fn add_oytlayer_to_linker<T, U>(
    linker: &mut Linker<T>,
    f: fn(&mut T) -> HostContext<&mut U>,
)
    where U: CryptoHost + 'static,
{
    add_to_linker::<_, OutlayerHost<U>>(linker, f);
}
```

However, in our current implementation we don't have any generic parameters for `HostContext` (InMemorySigner is used directly). So, we can pick a [simpler option](https://docs.rs/wasmtime/latest/wasmtime/component/trait.HasData.html#example-host-for-mylibrarystate-using-storet) for now:
```rust
pub fn add_my_library_to_linker<T>(
    linker: &mut Linker<T>,
    f: fn(&mut T) -> &mut HostContext,
) {
    // invoke the bindgen!-generated `add_to_linker`
    add_to_linker::<_, OutlayerHost>(linker, f);
}
``` 

So, for `vm-runner` crate we will have no generics for the underlying host impl - we use only our `HostContext<'static>`. If we ever come to generics (those will be needed for mocking) - we can introduce them following the first approach above

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+            code => Some(ExecutionError::NonZeroExit(code)),
+        };
+    }
+
+    let err = err
+        .downcast_ref::<Trap>()
+        .copied()
+        .map_or_else(|| ExecutionError::Unknown(err), ExecutionError::Trap);
+
+    tracing::debug!("Program trapped: {err}");
```
mitinarseny:
```suggestion
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,363 @@
...
+        &self,
+        ctx: Context<I, O, E, H>,
+        binary: impl AsRef<[u8]>,
+    ) -> Result<ExecutionOutcome>
+    where
+        I: StdinStream + 'static,
+        O: StdoutStream + 'static,
+        E: StdoutStream + 'static,
+    {
+        let component = self.compile(binary)?;
```
mitinarseny:
```suggestion
        let component = self.compile(binary).context("compile")?;
```

mitinarseny (changes requested):

crates/outlayer/cli/src/main.rs
hlgltvnnk: I don't like the idea of ​​using secrets as arguments to the utility - that's why it was put in env, but on the other hand, we could replace this with the path to the file where the secret is stored and then pass it as an argument, what do you think?

crates/outlayer/sdk/src/host/crypto/ed25519.rs
hlgltvnnk: it returns Vec<u8> from host function, updated - made in more concise

crates/outlayer/host/src/lib.rs
hlgltvnnk: May be make vm runner completely generic? Without any binding to defuse_outlayer_host - allow anything that implements the Linkable trait (create it)

crates/outlayer/vm-runner/src/outcome.rs
hlgltvnnk: Clone, Eq, PartialEq, Hash are not implemented for anyhow error

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: I think we should leave these limits - firstly, we are obliged to limit the maximum memory size for wasms - this way we can calculate the number of requests executed in parallel by single worker
secondly, the amount of fuel used is necessary for the execution result (which I would not remove - after all, it is convenient for clients, because it serves as an indicator for optimization) if we do not set this limit - it is impossible to apply the amount of gas used (runtime returns not how much was used, but how much is left)
I suggest simply adding docs into the cli that we have specified values by default if not set

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: Why might we want to set different limits for different wasms?

crates/outlayer/vm-runner/src/lib.rs
hlgltvnnk: I don't like the idea of ​​exposing the entire contents of files to a public API - developer can make a mistake and leave a certain object public, but it shouldn't be in the public API

crates/outlayer/vm-runner/src/context.rs
hlgltvnnk: renamed it to context and reexported it in lib.rs

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: We talked with you earlier that there may be a situation where we may want to mutate the host state data or read it without wrapping it over the store - this is what this example was created for

crates/outlayer/sdk/src/host/crypto/ed25519.rs
mitinarseny: it was already done in previous [PR](https://github.com/near/intents/pull/255) that is now merged

crates/outlayer/vm-runner/src/outcome.rs
mitinarseny: Then maybe let's add `ExectuionDetails`:
```rust
#[derive(Debug, Clone, Clone, Eq, PartialEq, Hash)]
pub struct ExecutionDetails {
    fiel_consumed: u64,
}
```
and use it in `ExecutionOutcome`:
```rust
#[derive(Debug)]
pub struct ExecutionOutcome {
    pub details: ExecutionDetails,
    pub error: Option<ExecutionError>,
}
```
So that callers can use it. Wdyt?

crates/outlayer/vm-runner/src/runtime.rs
mitinarseny: I get it, but it can be a decision of a caller of `vm-runner` crate whether to set these limits or not. For instance, TEE worker would want to set them, while CLI from this PR - not

crates/outlayer/vm-runner/src/runtime.rs
mitinarseny: E.g. because each execution requests pre-paid for different amounts of fuel/memory

crates/outlayer/vm-runner/src/lib.rs
mitinarseny: The same reasoning can be applied to `lib.rs` file. As developers, we need to use `pub(crate)` if this struct is intended to be re-used in other modules but not appear in public API

crates/outlayer/vm-runner/src/runtime.rs
mitinarseny: We should never mutate the data unless WASM execution succeeds. So, all host-function impls should not have any mutexes and we should not promote it to be used this way

crates/outlayer/sdk/src/host/crypto/ed25519.rs
hlgltvnnk: try into is chained to the second block - so while using this sdk u get an error - expected `[u8; 32]`, found `Vec<u8>`

```
    #[cfg(target_family = "wasm")]
    {
        sys::crypto::ed25519::derive_public_key(path.as_ref())
    }
    #[cfg(not(target_family = "wasm"))]
    {
        mock::HOST
            .with_borrow_mut(|h| h.derive_public_key(path.as_ref().to_string()))
            .expect("host")
    }
    .try_into()
    .expect("invalid length")
```

crates/outlayer/vm-runner/src/runtime.rs
hlgltvnnk: Unfortunately, given that we pass the fuel limit to the context (which means we can't know if the limit will ever be set), this meant that at initialization time we cant know whether to specify config.consume_fuel(true) in the engine. In case we specify it but don't set the fuel limit, the limit will be 0 - so in any case we will get OutOfFuel error
Because of this, we need to have a default value at the context level

hlgltvnnk: upd:  fuel limit eq u64::MAX in case if not set

crates/outlayer/host/src/lib.rs
```diff
@@ -7,18 +6,25 @@ pub use defuse_outlayer_crypto::signer::InMemorySigner;
 pub use defuse_outlayer_primitives as primitives;
 use defuse_outlayer_primitives::AppId;
 
+use crate::crypto::CryptoHost;
+
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Context<'a> {
+pub struct AppContext<'a> {
```
mitinarseny, line 11: Why `AppContext`? It might contain request-specific information, too. For example, `predecessor_id`, etc..
```suggestion
pub struct Context<'a> {
```

crates/outlayer/host/src/crypto/secp256k1.rs
```diff
@@ -1,10 +1,25 @@
...
 
 use crate::State;
 
-impl crate::bindings::outlayer::crypto::secp256k1::Host for State<'_> {
-    fn derive_public_key(&mut self, path: String) -> wasmtime::Result<Vec<u8>> {
+/// Trait defining secp256k1-related host functions available to the component
+pub trait Secp256k1Host: Send {
+    fn secp256k1_derive_public_key(&mut self, path: String) -> Result<Vec<u8>>;
+    fn secp256k1_sign(&mut self, path: String, msg: Vec<u8>) -> Result<Vec<u8>>;
+}
```
mitinarseny: I suggest we [don't](https://github.com/near/intents/pull/253#discussion_r3157552980) introduce generics for now, because they're very likely to be changed in the future. For example, we might want to make them `async` and/or to return specific error types.

We will figure it out later, but for now, please, remove these generic traits and keep generated bindings in this crate. Leave only single host implementation, configurable by `app_id` and `&InMemorySigner`.

crates/outlayer/host/Cargo.toml
```diff
@@ -13,8 +13,3 @@ defuse-outlayer-crypto = { workspace = true, features = ["ed25519", "secp256k1",
 defuse-outlayer-primitives = { workspace = true, features = ["digest"] }
 
 anyhow.workspace = true
-tracing.workspace = true
-wasmtime = { workspace = true, features = ["runtime", "component-model"] }
-
-[package.metadata.cargo-machete]
-ignored = ["tracing"] # required by `wasmtime::component::bindgen!()`
```
mitinarseny, line 20: Bring it back, please

crates/outlayer/vm-runner/src/bindings.rs
mitinarseny, line 1: bring it back please

crates/outlayer/vm-runner/cli/src/main.rs
```diff
@@ -0,0 +1,92 @@
+use anyhow::{Context, Result};
+use bytesize::ByteSize;
+use clap::Parser;
+use std::{borrow::Cow, path::PathBuf};
+
+const DEFAULT_APP_ID: &str = "near:0sab1c86e60758fe3e8fc7ae40ecd2df1a07513ca9";
```
mitinarseny: Can you please leave a comment on how it was generated?
Also, let's make it typed:
```suggestion
const DEFAULT_APP_ID: AppId<'static> =
    AppId::Near(Cow::Borrowed(
        AccountIdRef::new_or_panic("0sab1c86e60758fe3e8fc7ae40ecd2df1a07513ca9")
    ));
    
// then use it as `default_value_t = DEFAULT_APP_ID`
```

crates/outlayer/vm-runner-cli/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Path to a file containing the hex-encoded 32-byte seed for the host's signer key
+    seed_path: PathBuf,
+
+    /// Application ID to use in the host context
+    #[clap(long, short, default_value = DEFAULT_APP_ID)]
+    app_id: AppId<'static>,
+
+    /// Maximum number of WebAssembly instructions the component may execute
+    #[clap(long, short)]
+    fuel: Option<u64>,
```
mitinarseny: Will it be `u64::MAX` by default?

crates/outlayer/vm-runner-cli/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let args = Args::parse();
+
+    let seed_hex = std::fs::read_to_string(&args.seed_path)
+        .with_context(|| format!("failed to read seed file: {}", args.seed_path.display()))?;
+    let seed =
+        hex::decode(seed_hex.trim()).context("seed file must contain a hex-encoded byte string")?;
```
mitinarseny: Let's make this file already contain seed bytes, not hex-encoded

crates/outlayer/vm-runner-cli/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    let component = runner
+        .compile(&wasm_binary)
+        .with_context(|| format!("failed to compile: {}", args.wasm_path.display()))?;
+
+    let outcome = runner.execute(ctx, &component).await.context("execute")?;
+
+    outcome.into_result().context("component failed")?;
+
+    Ok(())
```
mitinarseny:
```suggestion
    outcome.into_result().context("component failed")
```

crates/outlayer/vm-runner/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+mod runtime;
+
+pub use defuse_outlayer_host as host;
+
+pub use wasmtime;
+pub use wasmtime_wasi;
+
+pub use error::ExecutionError;
+pub use outcome::{ExecutionDetails, ExecutionOutcome};
+pub use runtime::{ExecutionContext, VmRuntime};
```
mitinarseny:
```suggestion
pub use self::{
    error::ExecutionError,
    outcome::{ExecutionDetails, ExecutionOutcome},
    runtime::{ExecutionContext, VmRuntime},
};
```

crates/outlayer/vm-runner/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+mod bindings;
+mod context;
+mod error;
+mod outcome;
+mod runtime;
+
+pub use defuse_outlayer_host as host;
+
+pub use wasmtime;
+pub use wasmtime_wasi;
```
mitinarseny, line 11: Move re-exports to the bottom pls

crates/outlayer/vm-runner/src/outcome.rs
```diff
@@ -0,0 +1,32 @@
...
+#[non_exhaustive]
+pub struct ExecutionDetails {
+    pub fuel_consumed: u64,
+}
+
+impl ExecutionDetails {
+    pub const fn new(fuel_consumed: u64) -> Self {
+        Self { fuel_consumed }
+    }
+}
```
mitinarseny:
```suggestion
```

crates/outlayer/vm-runner/src/outcome.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// Outcome of executing a component in the VM runtime
+#[derive(Debug, Default)]
+#[non_exhaustive]
```
mitinarseny:
```suggestion
```

crates/outlayer/vm-runner/src/outcome.rs
```diff
@@ -0,0 +1,32 @@
...
+pub struct ExecutionOutcome {
+    pub details: ExecutionDetails,
+    pub error: Option<ExecutionError>,
+}
+
+impl ExecutionOutcome {
+    pub const fn new(details: ExecutionDetails, error: Option<ExecutionError>) -> Self {
+        Self { details, error }
+    }
```
mitinarseny:
```suggestion
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,288 @@
...
+};
+use wasmtime_wasi::{
+    WasiCtx,
+    cli::{StdinStream, StdoutStream},
+    p2::bindings::Command,
+};
+
+use crate::{bindings::Imports, outcome::ExecutionOutcome};
+use crate::{context::HostCtx, outcome::ExecutionDetails};
+use crate::{context::HostFunctionsImpl, error::ExecutionError};
```
mitinarseny: merge imports pls

crates/outlayer/vm-runner/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+workspace = true
+
+[dependencies]
+defuse-outlayer-host.workspace = true
+
+anyhow.workspace = true
+thiserror.workspace = true
+tracing.workspace = true
+
+wasmtime = { workspace = true, features = ["cranelift"] }
```
mitinarseny: Let's not enable this feature by default, this is dangerous given the number of hacks lately
```suggestion
wasmtime.workspace = true

[features]
cranelift = ["wasmtime/cranelift"]
```

Cargo.toml
```diff
@@ -48,8 +48,10 @@ members = [
   "crates/outlayer/crypto",
   "crates/outlayer/host",
   "crates/outlayer/primitives",
-  "crates/outlayer/sys",
   "crates/outlayer/sdk",
+  "crates/outlayer/sys",
+  "crates/outlayer/vm-runner",
+  "crates/outlayer/vm-runner-cli",
```
mitinarseny: Can we put `cli` inside of `vm-runner`?

```suggestion
  "crates/outlayer/vm-runner",
  "crates/outlayer/vm-runner/cli",
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,288 @@
...
+/// catch out-of-bounds accesses
+const MEMORY_GUARD_SIZE: u64 = 64 * 1024 * 1024; // 64 MiB
+
+/// Maximum fuel consumption allowed for a single component execution
+const MAX_FUEL_CONSUMPTION: u64 = u64::MAX;
+
+pub struct ExecutionContext<I, O, E, H> {
+    input: I,
+    output: O,
+    error: E,
```
mitinarseny:
```suggestion
    stdin: I,
    stdout: O,
    stderr: E,
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// Creates a new execution context
+    /// By default, there are no fuel or memory limits
+    #[must_use]
+    pub const fn new(input: I, output: O, error: E, host_state: H) -> Self {
+        Self {
+            input,
+            output,
+            error,
+            host_state,
+            fuel_limit: None,
```
mitinarseny: Since it turned out that fuel has to be set anyway, then let's not make it optional then but rather `u64::MAX` by default?

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,288 @@
...
+
+        let wasi_ctx = WasiCtx::builder()
+            .stdin(input)
+            .stdout(output)
+            .stderr(error)
+            .build();
+
+        let limits = memory_limit.map_or_else(StoreLimitsBuilder::new, |m| {
+            StoreLimitsBuilder::new().memory_size(m)
+        });
```
mitinarseny:
```suggestion
        let mut limits = StoreLimitsBuilder::new();
        if let Some(memory_limit) = memory_limit {
            limits = limits.memory_size(m);
        }
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,288 @@
...
+    H: HostFunctions,
+{
+    fn default() -> Self {
+        Self::new().expect("setup")
+    }
+}
+
+fn classify_error(err: anyhow::Error) -> Option<ExecutionError> {
+    if let Some(exit) = err.downcast_ref::<wasmtime_wasi::I32Exit>() {
+        tracing::debug!("Program exited with code {}", exit.0);
```
mitinarseny:
```suggestion
```

crates/outlayer/vm-runner/src/runtime.rs
```diff
@@ -0,0 +1,288 @@
...
+            0 => None,
+            code => Some(ExecutionError::NonZeroExit(code)),
+        };
+    }
+
+    Some(
+        err.downcast_ref::<Trap>()
+            .copied()
+            .map_or_else(|| ExecutionError::Unknown(err), ExecutionError::Trap),
+    )
```
mitinarseny:
```suggestion
    Some(
        err.downcast::<Trap>()
            .map_or_else(ExecutionError::Unknown, ExecutionError::Trap)
    )
```

crates/outlayer/host/src/crypto/secp256k1.rs
hlgltvnnk: 😢

crates/outlayer/vm-runner-cli/src/main.rs
hlgltvnnk: also add u64::MAX in Config constructor

crates/outlayer/vm-runner/Cargo.toml
hlgltvnnk: avoid Jit - > replace with interpretation

hlgltvnnk: postponed

Cargo.toml
```diff
@@ -111,6 +113,8 @@ defuse-outlayer-host.path = "crates/outlayer/host"
 defuse-outlayer-primitives.path = "crates/outlayer/primitives"
 defuse-outlayer-sdk.path = "crates/outlayer/sdk"
 defuse-outlayer-sys.path = "crates/outlayer/sys"
+defuse-outlayer-vm-runner.path = "crates/outlayer/vm-runner"
+defuse-outlayer-vm-runner-cli.path = "crates/outlayer/vm-runner-cli"
```
mitinarseny:
```suggestion
defuse-outlayer-vm-runner-cli.path = "crates/outlayer/vm-runner/cli"
```

crates/outlayer/vm-runner/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[lints]
+workspace = true
+
+[dependencies]
+defuse-outlayer-host.workspace = true
+
+anyhow.workspace = true
+thiserror.workspace = true
+tracing.workspace = true
```
mitinarseny, line 16: not needed for now?

mitinarseny (commented):
Aweomse! Let's make sure all previous unresolved comments are addresses and finally merge it! 🤞🏻

crates/outlayer/vm-runner/Cargo.toml
hlgltvnnk: used to instrument execution

crates/outlayer/vm-runner/src/runtime.rs
mitinarseny: didn't see `-> Result`, sorry

mitinarseny (dismissed):

PR #254, opened 2026-04-22
feat(outlayer): tower service 

pityjllk:
For now targetting `feat/outlayer-vm-runner` will switch to `main' after https://github.com/near/intents/pull/253 is merged.

Still need to do a bit of cleanup but 

``` 
cargo run -p defuse-outlayer-service --example run --  "data:application/wasm;base64,$(cat ~/near-pds.wasm| base64)" --wasm-hash 4d523c7ea691437c2af83cabf67102f5fb5a28fef8023d9a7da168fbc9fb0974 --string-input "hello"

SignedExecutionResponse {
    response: ExecutionResponse {
        request_id: 35f4d62e-e0e5-434f-8408-d18f779d1143,
        result: Ok(
            b"",
        ),
        metrics: ExecutionMetrics {
            instructions_used: 1246,
            wall_time: 0ns,
            compile_time: None,
        },
        storage: ProjectStorage,
        respond_to: OnChainDestination {
            contract_id: AccountId(
                "example.near",
            ),
        },
    },
    signature: ed25519::Signature {
        R: 0x1e66e2bfee2f3ec8d4ea3c098f73a3c7165961b5aebc4dc277f0605adc12dece,
        s: 0x59693d6c7b30614fabb467e3408fcdb059c9c85bfab1ce61038ba732162b4b09,
    },
}

```

## Release Notes

* **New Features**
  * Introduced a new outlayer service module enabling WASM component execution with configuration, caching, and retry mechanisms.
  * Added support for processing both on-chain and off-chain execution requests.
  * Integrated component resolution and response signing capabilities.

* **Tests**
  * Added integration tests for on-chain fetch service.

* **Chores**
  * Updated workspace configuration and dependencies to support the new service module.


crates/outlayer/service/examples/run.rs
```diff
@@ -0,0 +1,118 @@
...
+///   run '<https://example.com/component.wasm>' --wasm-hash <64 hex chars> < input.bin
+#[derive(Parser)]
+struct Args {
+    /// WASM URL — inline (`data:application/wasm;base64,…`) or remote (`http(s)://…`).
+    url: String,
+
+    /// SHA-256 of the WASM binary (64 hex chars).
+    #[arg(long)]
+    wasm_hash: String,
+
```
mitinarseny, line 34: Maybe calculate hash from downloaded file by default?

crates/outlayer/service/examples/run.rs
```diff
@@ -0,0 +1,118 @@
...
+/// Run a WASM component through the outlayer service stack.
+///
+/// Input is read from stdin; the component's stdout/stderr are forwarded to
+/// the host's stdout/stderr.
+///
+/// Examples:
+///   echo 'hello world' | run 'data:application/wasm;base64,AGFzbQ...' --wasm-hash <64 hex chars>
+///   run '<https://example.com/component.wasm>' --wasm-hash <64 hex chars> < input.bin
+#[derive(Parser)]
+struct Args {
```
mitinarseny, line 27: What about `app_id`?

crates/outlayer/service/examples/run.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// WASM URL — inline (`data:application/wasm;base64,…`) or remote (`http(s)://…`).
+    url: String,
```
mitinarseny, line 29:
```suggestion
    url: Url,
```

crates/outlayer/service/src/types.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// Simplified, internal execution request passed to `OutlayerService`.
+/// Contains only what WASM execution needs.
+#[derive(Debug, Clone)]
+pub struct ExecutionRequest {
+    /// Opaque identifier for this request, used for logging only.
+    pub request_id: String,
+    pub project_id: AccountId,
+    pub wasm_url: String,
```
mitinarseny, line 57:
```suggestion
    pub wasm_url: Url,
```

crates/outlayer/service/src/types.rs
```diff
@@ -0,0 +1,111 @@
...
+    OffChain(OffChainRequest),
+}
+
+/// Simplified, internal execution request passed to `OutlayerService`.
+/// Contains only what WASM execution needs.
+#[derive(Debug, Clone)]
+pub struct ExecutionRequest {
+    /// Opaque identifier for this request, used for logging only.
+    pub request_id: String,
+    pub project_id: AccountId,
```
mitinarseny, line 56: We call id `app_id: AppId`

crates/outlayer/service/src/types.rs
```diff
@@ -0,0 +1,111 @@
...
+    OnChain(OnChainRequest),
+    OffChain(OffChainRequest),
+}
+
+/// Simplified, internal execution request passed to `OutlayerService`.
+/// Contains only what WASM execution needs.
+#[derive(Debug, Clone)]
+pub struct ExecutionRequest {
+    /// Opaque identifier for this request, used for logging only.
+    pub request_id: String,
```
mitinarseny, line 55: If it's used only internally, then why not derive it internally and/or generate as UUID, for example?

crates/outlayer/service/src/types.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// Simplified, internal execution request passed to `OutlayerService`.
+/// Contains only what WASM execution needs.
+#[derive(Debug, Clone)]
+pub struct OffChainRequest {
+    /// Opaque identifier for this request, used for logging only.
+    pub request_id: String,
+    pub project_id: AccountId,
+    pub input: Bytes,
```
mitinarseny, line 40: What's the reason for using `Bytes`? Wouldn't `Vec<u8>` work?

crates/outlayer/service/examples/run.rs
mitinarseny: Can we accept `file://` URLs as well?

crates/outlayer/service/src/types.rs
```diff
@@ -0,0 +1,111 @@
...
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Deserialize)]
+pub struct OnChainRequest {
+    /// NOTE: its sitll not clear how to uniquely identify a request on chain. Lets
+    /// use placeholder for now ...
+    #[serde_as(as = "Hex")]
+    pub nonce: [u8; 32],
+    #[serde_as(as = "serde_with::base64::Base64")]
+    pub tx_hash: [u8; 32],
+    pub project_id: AccountId,
```
mitinarseny, line 25: Near tx hashes are encoded as base58

crates/outlayer/service/src/types.rs
```diff
@@ -0,0 +1,111 @@
+use bytes::Bytes;
+use serde::{Deserialize, Serialize};
+use serde_with::{hex::Hex, serde_as};
+
+/// Opaque NEAR account identifier. Wraps a validated string.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct AccountId(pub String);
```
mitinarseny, line 7: Why not using `near-account-id`?

crates/outlayer/service/src/types.rs
mitinarseny, line 1: Let's split things in different modules based on their semantics instead of types of Rust statements in it

crates/outlayer/service/examples/run.rs
```diff
@@ -0,0 +1,118 @@
...
+                project_id: req.project_id,
+                wasm_url: url,
+                wasm_hash,
+                input: req.input,
+            })
+        }
+    });
+
+    // Fixed key — replace with a real TEE key in production.
+    let signing_key = defuse_outlayer_crypto::signer::InMemorySigner::from_seed(&[1u8; 32]);
```
mitinarseny, line 88: Use the same seed as in SDK mock and vm-runner-cli. Maybe it makes sense to create `outlayer/test-utils` package where we can define it and re-use across different crates

crates/outlayer/service/examples/run.rs
```diff
@@ -0,0 +1,118 @@
...
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .init();
+
+    let args = Args::parse();
+
+    let wasm_hash = parse_hash_hex(&args.wasm_hash)?;
+    let wasm_url = args.url.clone();
+
+    let mut input = Vec::new();
+    std::io::stdin().read_to_end(&mut input)?;
```
mitinarseny, line 65: it would be nice if you can prepare everything to run before reading stdin, so that we get an error even if stdin is pending (e.g. waiting on tty)

mitinarseny: Let's use `io` from `tokio` instead of `std`

crates/outlayer/service/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+        .retry(Attempts(config.fetch.retry_attempts))
+        .timeout(config.fetch.timeout)
+        .service(storage::StorageFetchService);
+
+    // TODO: consider moving signing outside of the Tower stack
+    ServiceBuilder::new()
+        .map_err(timeout_err::<ExecutionStackError>)
+        .timeout(config.total_timeout)
+        .service(SigningService::new(
+            OutlayerService::new(executor, wasm, env, storage, on_chain),
```
mitinarseny, line 95: The comment is right: there should be no signing at this level. At least, signing should be "composable" with the underlying service, so we can use it only in context of a real TEE worker, while `service-cli` example would not sign anything: it simply doesn't have a key for that

crates/outlayer/service/src/executor.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn call(&mut self, req: WasmExecutionRequest) -> Self::Future {
+        let runtime = Arc::clone(&self.runtime);
+        let config = self.config.clone();
+        let host_state = self.host_template.clone();
+        Box::pin(
+            async move {
+                let stdout = MemoryOutputPipe::new(config.stdout_limit);
+                let stderr = MemoryOutputPipe::new(config.stderr_limit);
+                let ctx = vm_runner::Context::new(
+                    MemoryInputPipe::new(req.request.input),
```
mitinarseny, line 97: This is the only place where `req.request` is used. Moreover, only `input` field is used. This means that we have data duplication and we should avoid it at all costs, since it increases the likelyhood of making a mistake

crates/outlayer/service/src/types.rs
```diff
@@ -0,0 +1,111 @@
...
+/// Placeholder for on-chain project storage (key → value snapshot).
+#[derive(Debug, Clone, Default, Serialize)]
+pub struct ProjectStorage;
+
+/// Placeholder for decrypted project environment variables / secrets.
+#[derive(Debug, Clone, Default)]
+pub struct ProjectEnv;
+
+#[derive(Debug, Default, Serialize)]
+pub struct ExecutionMetrics {
```
mitinarseny, line 109: We have already this struct called `ExecutionDetails` in `vm-runner`

crates/outlayer/service/src/types.rs
```diff
@@ -0,0 +1,111 @@
...
+#[derive(Debug, Serialize)]
+pub struct ExecutionResponse {
+    // TODO: error string may be arbitrarily large; consider a dedicated SerializeAs that trims it
+    #[serde_as(as = "Result<serde_with::base64::Base64, serde_with::DisplayFromStr>")]
+    pub result: anyhow::Result<Bytes>,
+    #[serde_as(as = "serde_with::base64::Base64")]
+    pub logs: Bytes,
+
+    pub metrics: ExecutionMetrics,
+    pub storage: ProjectStorage,
```
mitinarseny, line 97: We don't have it for now, let's avoid unnecessary things

crates/outlayer/service/src/types.rs
```diff
@@ -0,0 +1,111 @@
...
+            wasm_url: r.wasm_url,
+            wasm_hash: r.wasm_hash,
+            input: r.input,
+        }
+    }
+}
+
+/// Raw output from `WasmExecutor`. Contains only execution results.
+#[serde_as]
+#[derive(Debug, Serialize)]
```
mitinarseny, line 88: We need to derive `Deserialize`, too. So this library can be re-used by clients.

mitinarseny: In fact, we need to split it into two crates, so that clients can depend only on API without the implementation itself

crates/outlayer/service/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+            async move {
+                tokio::task::spawn_blocking(move || {
+                    rt.compile(&*bytes).map_err(ExecutionStackError::Compile)
+                })
+                .await
+                .map_err(|e| ExecutionStackError::Compile(anyhow::anyhow!(e)))?
+            }
+        })
+        .map_err(timeout_err::<resolver::ResolveError>)
+        .retry(Attempts(config.fetch.retry_attempts))
```
mitinarseny, line 72: How about exponential backoff?

crates/outlayer/service/src/config.rs
```diff
@@ -0,0 +1,38 @@
...
+        Self {
+            timeout: Duration::from_secs(10),
+            retry_attempts: 2,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Config {
+    pub total_timeout: Duration,
```
mitinarseny, line 23: derive serde?

crates/outlayer/service/src/executor.rs
```diff
@@ -0,0 +1,134 @@
...
+                    .map_or_else(|| Ok(stdout_bytes), |e| Err(anyhow::Error::from(e)));
+
+                Ok(ExecutionResponse {
+                    result,
+                    logs: stderr_bytes,
+                    metrics: ExecutionMetrics { instructions_used },
+                    storage: ProjectStorage,
+                })
+            }
+            .instrument(tracing::Span::current()),
```
mitinarseny, line 131: I guess, there should be already existing tracing instrumentation layer for tower

crates/outlayer/service/src/executor.rs
mitinarseny: Are there any limits on the length of stdin in one of the top-level Layer's?

crates/outlayer/service/src/outlayer_app_client.rs
```diff
@@ -0,0 +1,51 @@
+use near_api::{Contract, NetworkConfig};
```
mitinarseny, line 1: Please, use `near-kit-rs` instead

crates/outlayer/service/src/outlayer_app_client.rs
```diff
@@ -0,0 +1,51 @@
...
+        Self {
+            account_id,
+            network_config,
+        }
+    }
+
+    // TODO: use generated bindings from contract ABI instead of manual JSON deserialization
+    pub async fn oa_code_hash(&self) -> Result<[u8; 32], OutlayerAppClientError> {
+        let hash_hex: String = Contract(self.account_id.clone())
+            .call_function("oa_code_hash", ())
```
mitinarseny, line 28: Use `#[near_kit::contract]` macro insead

crates/outlayer/service/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+workspace = true
+
+[dependencies]
+
+defuse-outlayer-crypto = { workspace = true, features = ["ed25519", "signing"] }
+defuse-outlayer-host = { workspace = true }
+defuse-outlayer-vm-runner = { workspace = true }
+
+near-api.workspace = true
+near-sdk.workspace = true
```
mitinarseny, line 18: Avoid any dependencies on `near-sdk`

crates/outlayer/service/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+
+[lints]
+workspace = true
+
+[dependencies]
+
+defuse-outlayer-crypto = { workspace = true, features = ["ed25519", "signing"] }
+defuse-outlayer-host = { workspace = true }
+defuse-outlayer-vm-runner = { workspace = true }
+
```
mitinarseny, line 16: `vm-runner` already re-exports `defuse_outlayer_host as host`
```suggestion
defuse-outlayer-vm-runner.workspace = true

```

crates/outlayer/service/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+version.workspace = true
+rust-version.workspace = true
+repository.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+
+defuse-outlayer-crypto = { workspace = true, features = ["ed25519", "signing"] }
```
mitinarseny, line 13: Instead of depending on it here, let's re-export it from `defuse-outlayer-host`

crates/outlayer/service/src/utils/cache.rs
```diff
@@ -0,0 +1,113 @@
...
+            0 // never called
+        }
+    }
+    let mut h = Sha256Hasher(Sha256::new());
+    key.hash(&mut h);
+    h.0.finalize().into()
+}
+
+#[derive(Clone)]
+pub struct CacheLayer<K, V> {
```
mitinarseny, line 44:
```suggestion
pub struct InMemoryLruCache<K, V> {
```

crates/outlayer/service/src/utils/cache.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn default() -> Self {
+        Self {
+            capacity: NonZeroUsize::new(100).unwrap(),
+            max_fetch_bytes: 100 * 1024 * 1024, // 100 MiB
+        }
+    }
+}
+
+fn cache_key<K: Hash>(key: &K) -> [u8; 32] {
+    struct Sha256Hasher(Sha256);
```
mitinarseny, line 29: let's make it reusable and generic over `Digest` and put it in a separate `hash-utils` crate?

crates/outlayer/service/src/utils/cache.rs
```diff
@@ -0,0 +1,113 @@
...
+impl Default for CacheConfig {
+    fn default() -> Self {
+        Self {
+            capacity: NonZeroUsize::new(100).unwrap(),
+            max_fetch_bytes: 100 * 1024 * 1024, // 100 MiB
+        }
+    }
+}
+
+fn cache_key<K: Hash>(key: &K) -> [u8; 32] {
```
mitinarseny, line 28: Are use sure that `Vec` it its `Hash` impl doesn't prepend its length?

crates/outlayer/service/src/executor.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl<H: HostFunctions + Clone + Send + 'static> Service<WasmExecutionRequest> for WasmExecutor<H> {
+    type Response = ExecutionResponse;
+    type Error = WasmEnvironmentInternalError;
+    type Future = BoxFuture<'static, Result<ExecutionResponse, WasmEnvironmentInternalError>>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
```
mitinarseny, line 85: We need to have configurable limits on how many concurrent executions we can make

crates/outlayer/service/src/utils/retry.rs
```diff
@@ -0,0 +1,22 @@
+use std::future::Ready;
+
+use tower::retry::Policy;
+
+#[derive(Clone)]
+pub struct Attempts(pub usize);
```
mitinarseny, line 6: This, and backoff impls already exist in `tower::retry`

mitinarseny (changes requested):
Great job, but we need to refactor it massively:
* remove/reuse existing abstractions
* remove everything connected to WASM storage and other environment
* etc...

crates/outlayer/service/src/executor.rs
```diff
@@ -0,0 +1,134 @@
...
+
+use crate::types::{
+    AccountId, ExecutionMetrics, ExecutionRequest, ExecutionResponse, ProjectEnv, ProjectStorage,
+};
+
+#[derive(Debug, Clone)]
+pub struct WasmExecutorConfig {
+    pub stdout_limit: usize,
+    pub stderr_limit: usize,
+    pub fuel_limit: u64,
```
hlgltvnnk: May be also add memory limit?

mitinarseny:
Closing in favor of #262. Maybe we would want to re-introduce tower on a high-level afterwards

PR #255, opened 2026-04-27
feat(outlayer): crypto, SDK, mock

mitinarseny:
This PR adds implementations for cryptography-related host-functions, enriches SDK and adds mocking capabilities for testing. As for now, it's only possible to override mocked host via environment variables 

## Summary by CodeRabbit

* **New Features**
  * Ed25519 and Secp256k1 support with non‑hierarchical key derivation, an in‑memory signer, AppId and DerivationPath types, and an SDK example for end‑to‑end signing.
  * SDK now exposes common encoding/serde helpers and re-exports primitives for downstream use.

* **Breaking Changes**
  * Secp256k1 signing now requires a 32‑byte prehash; host APIs return fixed-size byte arrays/vectors and host component surface changed.

* **Chores**
  * Updated cryptography and serialization library versions.


crates/outlayer/crypto/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// some curves can sign arbitrary byte slices, while others might expect
+    /// prehashed messages of a specific size.
+    type Message: ?Sized;
+
+    /// Signature of the curve
+    type Signature;
+
+    /// Derive curve-specific [tweak](Self::Tweak) from a **uniform**
+    /// digest.
+    fn tweak(hash: [u8; 32]) -> Self::Tweak;
```
pityjllk: everything is very generic but this tweak method... it feels like it doesnt not have to be part of this trait.

crates/outlayer/sdk/Cargo.toml
```diff
@@ -9,10 +9,32 @@ repository.workspace = true
...
 
 [target.'cfg(target_family = "wasm")'.dependencies]
 defuse-outlayer-sys.workspace = true
 
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
-defuse-outlayer-worker-host.workspace = true
+defuse-outlayer-host.workspace = true
+
+anyhow.workspace = true
+config = { workspace = true, features = ["json"] }
```
pityjllk: this feels a bit like overkill to me.... serde, config, json parsing. i guess having set fn that alows you to inject particular signer and maybe `fixture` to use default/custom one would be more ergonomic, otherwise you need to run tests in separate process with custom env variables to get it to work right 🤔 ?

pityjllk (commented):
lgtm, 2 minor nits

crates/outlayer/crypto/src/signer.rs
```diff
@@ -0,0 +1,136 @@
...
+
+impl InMemorySigner {
+    const HKDF_SEED_SALT: &'static [u8] = b"outlayer v0.1.0 signer seed:";
+
+    #[cfg(feature = "ed25519")]
+    const HKDF_INFO_ED25519_ROOT_SK: &'static [u8] = b"ed25519/root_sk";
+    #[cfg(feature = "secp256k1")]
+    const HKDF_INFO_SECP256K1_ROOT_SK: &'static [u8] = b"secp256k1/root_sk";
+
+    pub fn from_seed(seed: &[u8]) -> Self {
```
hlgltvnnk, line 30: In my opinion, we should specify the size of the seed - since salt and info are public, the security of the system depends entirely on the entropy of the seed

crates/outlayer/crypto/src/ed25519.rs
```diff
@@ -0,0 +1,118 @@
...
+                // reuse the same nonce for different challenges, as it might
+                // lead to leaking the root private key.
+                //
+                // Here we generate the hash, and thus, the nonce randomly.
+                // As a result, signature will be different every time, even
+                // if the same message is singed with the same tweak.
+                //
+                // TODO: derive hash_prefix deterministically from
+                // `root_sk.hash_prefix` and `tweak`?
+                hash_prefix: UnwrapErr::<SysRng>::default().random(),
```
hlgltvnnk, line 74: I don't like the idea of ​​using a random value for the nonce because:
1. we depend on the entropy of sys rng
2. Repetitions of hash_prefix are theoretically possible
3. non-consensual - shold be  same key + same message = deterministic signature
What do you think about deterministic nonce? for example hash with scalar + tweak

crates/outlayer/crypto/src/lib.rs
mitinarseny: Note that deriving does not take &self: such design enforces non-hardened public key derivation, so that a caller must use the same tweak for deriving public key and signing.

pityjllk (dismissed):
lgtm

hlgltvnnk (dismissed):

crates/outlayer/crypto/src/secp256k1.rs
```diff
@@ -0,0 +1,151 @@
...
+    type Path = [u8; 32];
+    type PublicKey = VerifyingKey;
+    /// Prehash, i.e. output of a cryptographic hash function
+    type Message = [u8; 32];
+    type Signature = (Signature, RecoveryId);
+
+    fn verify(
+        public_key: &VerifyingKey,
+        prehash: &[u8; 32],
+        (signature, _recovery_id): &Self::Signature,
```
gilcu3, line 33: why do you need to use prehash instead of the original message?

And why having the recovery_id if it is not used?

crates/outlayer/crypto/src/ed25519.rs
```diff
@@ -0,0 +1,131 @@
...
+pub use ed25519_dalek::{self, Signature, VerifyingKey};
+
+use crate::{DerivableCurve, DerivablePublicKey};
+
+pub struct Ed25519;
+
+impl Ed25519 {
+    fn tweak(path: &[u8; 32]) -> Scalar {
+        // TODO: are we sure there is no need to clamp?
+        Scalar::from_bytes_mod_order(*path)
```
gilcu3, line 13: can you expand on this comment? Why would we need to clamp here?

crates/outlayer/crypto/src/ed25519.rs
```diff
@@ -0,0 +1,131 @@
...
+    }
```
gilcu3, line 14: what does the tweak represent in your system? Notice that for mpc the tweak is the result of a hash function:

https://github.com/near/mpc/blob/e77356803634c88602ad458ddc17fbceff08c735/crates/near-mpc-crypto-types/src/kdf.rs#L10-L13

and that has some security implications, depending on your attacker model

crates/outlayer/crypto/src/ed25519.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl DerivableCurve for Ed25519 {
+    type Path = [u8; 32];
+    type PublicKey = VerifyingKey;
+    type Message = [u8];
+    type Signature = Signature;
+
+    fn verify(public_key: &VerifyingKey, msg: &[u8], signature: &Signature) -> bool {
+        public_key.verify_strict(msg, signature).is_ok()
+    }
```
gilcu3, line 25: can you guarantee that the signatures arriving here will always pass the strict verification if generated honestly?

crates/outlayer/crypto/src/ed25519.rs
```diff
@@ -0,0 +1,131 @@
...
+                // lead to leaking the root private key.
+                //
+                // Here we generate the hash, and thus, the nonce randomly.
+                // As a result, signature will be different every time, even
+                // if the same message is singed with the same tweak.
+                //
+                // TODO: derive hash_prefix deterministically from
+                // `root_sk.hash_prefix` and `tweak`?
+                hash_prefix: UnwrapErr::<SysRng>::default().random(),
+            };
```
gilcu3, line 75: it would be better to have determinism here, like it is proposed in the TODO

crates/outlayer/primitives/src/crypto.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[cfg(feature = "digest")]
+const _: () = {
+    use digest::{Digest, Output, Update};
+    use digest_io::IoWrapper;
+    use sha3::Sha3_256;
+
+    impl DerivationPath<'_> {
+        // TODO: do we need other domain separators?
+        const PREFIX: &'static [u8] = b"outlayer v0.1.0 tweak derivation:";
```
gilcu3, line 33: are you reusing them for ECDSA and ed25519? then probably yes it would be theoretically better to use different ones

crates/outlayer/primitives/src/crypto.rs
```diff
@@ -0,0 +1,85 @@
...
+
+        fn digest<D>(&self) -> Output<D>
+        where
+            D: Digest + Update,
+        {
+            let mut hasher = IoWrapper(D::new_with_prefix(Self::PREFIX));
+            borsh::to_writer(&mut hasher, self).expect("borsh");
+            hasher.0.finalize()
+        }
+    }
```
gilcu3, line 49: I see, so this is what later becomes `path` in the crypto key derivation. Would be nice to call it a differently, to avoid confusion

crates/outlayer/primitives/src/crypto.rs
```diff
@@ -0,0 +1,85 @@
...
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    cfg_attr(feature = "abi", derive(schemars::JsonSchema))
+)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+/// Total **non-hierarchical** derivation path
+pub struct DerivationPath<'a> {
+    /// Identifier of an application to derive for
+    pub app_id: AppId<'a>,
```
gilcu3, line 20: not sure where should this be mention, but the whole security of the per-app keys depends on these `app_id` being unique and authenticated, so that no app can pretend to be another one

crates/outlayer/crypto/src/signer.rs
```diff
@@ -0,0 +1,142 @@
...
+                    .expect("HKDF: ed25519");
+                crate::ed25519::SigningKey::from_bytes(&sk)
+            },
+            #[cfg(feature = "secp256k1")]
+            secp256k1_root_sk: {
+                let mut sk = [0u8; 32];
+                hk.expand(Self::HKDF_INFO_SECP256K1_ROOT_SK, &mut sk)
+                    .expect("HKDF: expand");
+                crate::secp256k1::SigningKey::from_bytes(&sk.into())
+                    .expect("secp256k1: zero scalar")
```
gilcu3, line 47: nit: I think this might fail in other cases (not only zero scalar), with negligible probability

crates/outlayer/crypto/src/secp256k1.rs
mitinarseny: > why do you need to use prehash instead

Prehash here lets you use different hash functions (as long as their output is 32 bytes). E.g. EVM txs and ERC-191 off-chain messages need prehash to be output of Keccak256, while other standards might to use SHA-256 or SHA3-256, for example.

AFAIU, the same approach is [used](https://github.com/near/mpc/blob/1f833a13f70addc34eb1cff704f93fec61e7f7eb/crates/near-mpc-crypto-types/src/primitives.rs#L99) by Near MPC contract

> And why having the recovery_id if it is not used?

This is because the `Curve::verify` trait method requires public key to be passed, so we don't need to restore it. If needed, one can use `VerifyingKey::recover_from_prehash` method directly.

gilcu3: yes we use the same, but I have no idea what's the use case here. Do you have any doc I can read about that?

crates/outlayer/crypto/src/ed25519.rs
mitinarseny: Basically, `ed25519-dalek` [does](https://github.com/dalek-cryptography/curve25519-dalek/blob/5b796f38c805311b296450f73a644732438d8ef9/ed25519-dalek/src/hazmat.rs#L91-L93) the [clamping](https://neilmadden.blog/2020/05/28/whats-the-curve25519-clamping-all-about/) in its `ExpandedSecretKey::from_bytes` implementation. So, I'm wondering, should we do the same here?

crates/outlayer/crypto/src/ed25519.rs
mitinarseny: Yes, we [use](https://github.com/near/intents/blob/1823243edc5f8093ec1225393d35854251d2ef07/crates/outlayer/primitives/src/crypto.rs#L35-L38) output of SHA3-256 over borsh-serialized `DerivationPath { app_id, path }`, prefixed with `outlayer v0.1.0 tweak derivation:`

crates/outlayer/crypto/src/ed25519.rs
mitinarseny: Can you clarify the question a bit please?

crates/outlayer/crypto/src/ed25519.rs
mitinarseny: What exact algorithm would you suggest for deriving `derived_esk.hash_prefix` here?

crates/outlayer/primitives/src/crypto.rs
mitinarseny: I was also leaning towards it but then couldn't come up with a clear answer why. Can you please share what exactly makes you think it would be better to use different prefixes for different curves?

crates/outlayer/primitives/src/crypto.rs
mitinarseny: Yes, we enforce it by deterministically deriving them from StateInit of [OutlayerApp](https://github.com/near/intents/blob/1823243edc5f8093ec1225393d35854251d2ef07/contracts/outlayer-app/README.md) contract. Or by getting them from on-chain view-calls to  contract to Near RPC in a sidecar container, all executing in TEE together.

crates/outlayer/crypto/src/signer.rs
mitinarseny: > this might fail in other cases

Hmm, which other cases are you referring to?

crates/outlayer/crypto/src/secp256k1.rs
mitinarseny: This is then exposed as a [host function](https://github.com/near/intents/blob/1823243edc5f8093ec1225393d35854251d2ef07/crates/outlayer/sdk/src/host/crypto/secp256k1.rs#L41-L62) to WASMs, so that they can sign arbitrary prehash being an output of cryptographic hash function depending on the exact signing scheme being used. Here is an example [showcase](https://github.com/near/intents/blob/1823243edc5f8093ec1225393d35854251d2ef07/crates/outlayer/sdk/examples/signer.rs) of using this API (not for production purposes, obviously)

crates/outlayer/crypto/src/ed25519.rs
gilcu3: to phrase it differently, is this verification function expected to be used only to verify signatures generated by this crate?

mitinarseny: No, signature verification are not yet added as host-functions. Nevertheless, we might want to add it in the future. Would you suggest to drop strictness requirement? What are the example use-cases where strictness might not be guaranteed or, even, cannot be achieved at all? E.g. would ed25519 MPC signatures pass this check?

crates/outlayer/crypto/src/ed25519.rs
gilcu3: I believe the `clamping` there fulfills another purpose, albeit similar. If you want to use 32 bytes as a private key for `ed25519` then you would do clamping (as explained in the nice [blog](https://neilmadden.blog/2020/05/28/whats-the-curve25519-clamping-all-about/) you shared) as a cheap way to make it nice (in the right range, avoiding small sub-groups). Here we are anyway adding the tweak to the private key, so clamping the tweak does not seem to help.

crates/outlayer/primitives/src/crypto.rs
gilcu3: I think it is mainly to avoid algebraic relations between derived keys that would otherwise not exist. In practice I don't think there is any reason to worry here, but adding "more" domain separators as in the comment is the usual recommendation.

crates/outlayer/crypto/src/signer.rs
gilcu3: if you check the call path starting in `SigningKey::from_bytes` you will arrive at this piece of code:

```
/// Create a new scalar from [`Curve::Uint`].
    pub fn new(uint: C::Uint) -> CtOption<Self> {
        CtOption::new(Self { inner: uint }, uint.ct_lt(&Self::MODULUS))
    }
```
it was also pointed out by an AI [above](https://github.com/near/intents/pull/255#discussion_r3144601566) :)

crates/outlayer/crypto/src/ed25519.rs
gilcu3: > E.g. would ed25519 MPC signatures pass this check?

that is a very good example. If you see our [tests](https://github.com/near/mpc/blob/main/crates/threshold-signatures/src/frost/eddsa/sign.rs) for FROST, we don't use the strict variant. So it might be that it does not work.

Also notice in the function docs:

https://docs.rs/ed25519-dalek/latest/ed25519_dalek/struct.VerifyingKey.html#method.verify_strict

> This version of verification is technically non-RFC8032 compliant.

crates/outlayer/crypto/src/ed25519.rs
gilcu3: This is not a simple question, but afaik to be as close as possible to RFC 8032 and keep the determinism property, something like:

```rust
let derived_hash_prefix: [u8; 32] = Sha512::new()
    .chain_update(b"outlayer/ed25519/derived-prefix/v1") // just another domain separator constant
    .chain_update(root_esk.hash_prefix)
    .chain_update(tweak.to_bytes())
    .finalize()[..32]
    .try_into()
    .expect("SHA-512 output is 64 bytes");

let derived_esk = ExpandedSecretKey {
    scalar: root_esk.scalar + tweak,
    hash_prefix: derived_hash_prefix,
};
```
would work.

Internally in `raw_sign`, the nonce will depend on the `msg` in any case.

crates/outlayer/crypto/src/ed25519.rs
mitinarseny: Thanks for flagging it! Will fix then

crates/outlayer/crypto/src/ed25519.rs
mitinarseny: I see, thanks!

> ```rust
>     .chain_update(tweak.to_bytes())
> ```

Would it be better to use the initial [`path: &[u8; 32]`](https://github.com/near/intents/blob/1823243edc5f8093ec1225393d35854251d2ef07/crates/outlayer/crypto/src/ed25519.rs#L54) here instead of [converted](https://github.com/near/intents/blob/1823243edc5f8093ec1225393d35854251d2ef07/crates/outlayer/crypto/src/ed25519.rs#L13) Scalar? I guess, the latter might have smaller entropy due to modular [reduction](https://github.com/dalek-cryptography/curve25519-dalek/blob/5b796f38c805311b296450f73a644732438d8ef9/curve25519-dalek/src/scalar.rs#L241-L242). Does it make sense to you?

gilcu3: >Does it make sense to you?

Yeah it makes sense, although I am not sure if it would mess up some API you already have

mitinarseny: Cool, thanks!
Final question: since `hash_prefix` is, surprisingly, [used](https://github.com/dalek-cryptography/curve25519-dalek/blob/5b796f38c805311b296450f73a644732438d8ef9/ed25519-dalek/src/signing.rs#L889) as a hash prefix for deterministic nonce derivation, isn't it an overkill derive it here by using hash/HKDF again? Wouldn't simple XOR'ing `root_esk.hash_prefix ^ path` suffice here?

gilcu3: That sounds a bit dangerous, although I cannot really say why from the back of my head. `path` is fully user controlled, which might have some implication.

PR #260, opened 2026-05-05
feat(outlayer): Component cache

pityjllk:


crates/outlayer/executor/src/compiler.rs
```diff
@@ -0,0 +1,21 @@
...
+
+#[derive(Clone)]
+pub struct Compiler(Arc<VmRuntime>);
+
+impl Compiler {
+    pub(super) const fn new(runtime: Arc<VmRuntime>) -> Self {
+        Self(runtime)
+    }
+
+    pub fn compile(&self, wasm: Bytes) -> Result<Component, CompileError> {
```
mitinarseny: There can be a support for WAT format, so let's keep the interface the same as in downstream:
```suggestion
    pub fn compile(&self, wasm: impl AsRef<[u8]>) -> Result<Component, CompileError> {
```

crates/outlayer/executor/src/compiler.rs
```diff
@@ -0,0 +1,21 @@
...
+        Ok(self.0.compile(wasm)?)
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+#[error(transparent)]
+pub struct CompileError(#[from] pub anyhow::Error);
```
mitinarseny: why do you need this newtype?

crates/outlayer/service/src/resolver/url/http.rs
```diff
@@ -2,6 +2,7 @@ use bytes::Bytes;
 use reqwest::{Client, Result};
 use url::Url;
 
+#[derive(Clone)]
 pub struct HttpResolver {
     // TODO: caching?
     client: Client,
```
mitinarseny, line 8: If you want to make it clonable, then let's wrap inner http `Client` in `Arc`, so that cloning it doesn't lose the cached connections, etc.. Or, otherwise, let's not make it clonable at all, so it will force callers to wrap it in `Arc`, too. Since can be cache in the future, I'm more in favor of the second approach

crates/outlayer/service/src/resolver/url/mod.rs
```diff
@@ -6,6 +6,7 @@ use url::Url;
 
 use self::http::HttpResolver;
 
+#[derive(Clone)]
 pub struct UrlResolver {
     http: HttpResolver,
```
mitinarseny, line 11: [Same](https://github.com/near/intents/pull/260/changes#r3190682699)

crates/outlayer/service/src/code.rs
```diff
@@ -66,8 +59,19 @@ pub struct AppCodeUrl {
...
+impl From<Bytes> for AppCodeUrl {
+    fn from(code: Bytes) -> Self {
+        Self {
+            code_url: format!("data:application/wasm;base64,{}", URL_SAFE.encode(&code))
+                .parse()
+                .expect("URL: parse"),
+            code_hash: Sha256::digest(&code).into(),
+        }
+    }
+}
```
mitinarseny: Let's make it a `from_code` method instead

crates/outlayer/service/src/code.rs
```diff
@@ -66,8 +59,19 @@ pub struct AppCodeUrl {
 }
 
 impl AppCodeUrl {
-    pub fn app_id(&self) -> AppId<'static> {
+    pub fn immutable_app_id(&self) -> AppId<'static> {
         // TODO: derive from state_init
         todo!()
     }
 }
```
mitinarseny, line 76: Why should it matter for callers if it's mutable or not?

crates/outlayer/service/src/lib.rs
```diff
@@ -1,49 +1,96 @@
...
+use defuse_outlayer_executor::{
+    self as executor, CompileError, Component, Context, Executor, HostContext, Outcome,
+};
+use moka::future::Cache;
 
+#[derive(Clone)]
 pub struct Outlayer {
     resolver: Resolver,
     executor: Executor,
+    runtime_cache: Option<Cache<Bytes, Component>>,
```
mitinarseny: I guess, the whole idea was not to store raw WASM `Bytes`, but rather SHA256 of it mapped to a compiled component, no?

crates/outlayer/service/src/lib.rs
```diff
@@ -1,49 +1,96 @@
...
+        cache: Option<Cache<Bytes, Component>>,
+    ) -> Self {
+        Self {
+            resolver,
+            executor,
+            runtime_cache: cache,
+        }
+    }
+
+    async fn resolve(&self, app: Code<'_>) -> Result<(HostContext<'static>, Bytes), Error> {
```
mitinarseny: `HostContext` can depend not only on `app_id`: in the future, there will be `predecessor_id`, too. So, I guess, this method should return `(AppId<'static>, Bytes)`, while `HostContext` is constructed on a higher level

crates/outlayer/service/src/lib.rs
mitinarseny: And why is `runtime_cache` optional?

crates/outlayer/service/src/lib.rs
```diff
@@ -1,49 +1,96 @@
...
+    runtime_cache: Option<Cache<Bytes, Component>>,
 }
 
 impl Outlayer {
-    pub async fn execute(&self, app: Code<'_>, input: Bytes, fuel: u64) -> Result<Outcome, Error> {
-        let app_id = app.app_id();
+    pub const fn new(
+        resolver: Resolver,
+        executor: Executor,
+        cache: Option<Cache<Bytes, Component>>,
```
mitinarseny: This constructor leaks abstraction. I suggest we create them inside instead

crates/outlayer/service/Cargo.toml
```diff
@@ -18,10 +18,12 @@ base64.workspace = true
 bytes.workspace = true
 data-url.workspace = true
 futures.workspace = true
+moka = { workspace = true, features = ["future"] }
```
mitinarseny, line 22: Why do you need async version of it? Compilation is synchronous

crates/outlayer/service/Cargo.toml
```diff
@@ -18,10 +18,12 @@ base64.workspace = true
...
 near-kit.workspace = true
 reqwest.workspace = true
 sha2.workspace = true
 thiserror.workspace = true
+tokio = { workspace = true, features = ["rt"] }
```
mitinarseny, line 27: Why do you need `tokio` here?

mitinarseny (changes requested):

crates/outlayer/executor/src/compiler.rs
pityjllk: yeah, thats probably can be removed

crates/outlayer/service/src/resolver/url/http.rs
pityjllk: fyi: Client already wraps its internals in Arc

pityjllk: > Since can be cache in the future, I'm more in favor of the second approach

not sure if i got this part

crates/outlayer/service/src/resolver/url/mod.rs
pityjllk: https://github.com/near/intents/pull/260#discussion_r3193914529

since we will most likely go with tower we will  clone 'defuse_outlayer_service::Outlayer` per incomming connection and wrap it in service function. In that case it would be nice to reuse things like compiled component cache, connection pool between instances of `Outlayer` following the reasoning from `reqwest::Client` where your instance is `Clone` and uses `Arc` internally and effectively  reuses some shared state. Basicaly i see no reason why not to do it in our case , wdyt?

crates/outlayer/service/src/code.rs
pityjllk: i fall for the case where

```
 CodeRef::AppId(app_id)
```

then you can do 

```
  let u: AppCodeUrl = resolver.resolve_app_id(app_id).await?;
  assert_eq!(u.app_id(), app_id); // FAILS
```

it was very misleading to me. basically having AppCodeUrl you can derive app_id but only from state_init where you need to assume some `admin_id`. My understanding is that its gonna be something like 0 address and will be used mainly for the purpose of referencing immutable wasms,  🤔 

maybe it could be renamed but for the reason of above roundtrip example i would use a different name than just `app_id` that sort of hides the fact that there is state_init behind with some hardcoded `admin_id`

crates/outlayer/service/src/lib.rs
pityjllk: > I guess, the whole idea was not to store raw WASM Bytes, but rather SHA256 of it mapped to a compiled component, no?

i had some second thoughs, cause ideally you should Eq guarantees of in case of cache collision but i guess i was overthinking it 🫠

pityjllk: > And why is runtime_cache optional?

mainly for testing, but seems like you can still configure cache_size == 0 so ill get rid of `Option`

crates/outlayer/service/src/lib.rs
pityjllk: went with builder pattern

crates/outlayer/service/Cargo.toml
pityjllk: compilation is synchronous but also cpu heavy task, so it should not block the tokio runtime. Thefore should be run in thread dedicated for cpu heavy operations which effectively makes it async operation. 

Also concurent calls to "try_get_with" would effectively lock the tokio runtime

crates/outlayer/service/Cargo.toml
pityjllk: https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

https://github.com/near/intents/pull/260#discussion_r3194769016

mitinarseny (approved):

PR #263, opened 2026-05-06
fix!(wallet): idempotent ops

mitinarseny:
This PR makes all current ops to have idempotent semantics

## Summary by CodeRabbit

* **Bug Fixes**
  * Wallet operations (`SetSignatureMode`, `AddExtension`, `RemoveExtension`) are now idempotent—repeating the same operation returns success instead of an error.
  * Events are only emitted when actual state changes occur, reducing unnecessary event noise.

* **Removals**
  * Custom wallet operations are no longer supported.


mitinarseny:
closing in favor of #303

PR #272, opened 2026-05-16
feat(kdf): composable signing and offline derivation

mitinarseny:
## Summary by CodeRabbit

* **New Features**
  * Full KDF support for Ed25519 and Secp256k1, including MPC-aware and app-scoped derivation for stable, application-specific keys.
  * New in-memory signer providing deterministic seed-derived root keys and path-based signing; CLI and runtime now use this signer plumbing.

* **Refactor**
  * Crypto and host layers reworked to a pluggable signer interface with unified KDF/signer APIs and modularized derivation/schema components for easier extension and testing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/near/intents/pull/272?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


crates/kdf/src/schema/additive.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// Multiply given scalar by a generator (i.e. base) point of the curve.
+    fn mul_by_generator(scalar: &Self::Scalar) -> Self::Point;
+
+    /// Convert public key into a point
+    fn pk2point(public_key: &Self::PublicKey) -> Self::Point;
+
+    /// Convert a point to a public key
+    fn point2pk(point: Self::Point) -> Self::PublicKey;
+}
```
pityjllk, line 67: move at the top of the file

crates/kdf/src/schema/additive.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        // pk' <- pk + G * tweak
+        let derived_point = master_pk + C::mul_by_generator(&tweak);
+
+        C::point2pk(derived_point)
+    }
+}
+
+/// Curve arithmetics used by [`Additive`] schema.
+pub trait CurveArithmetic: Curve {
```
pityjllk, line 52:
```suggestion
pub trait AdditiveCruve: Curve {
```
i dont like this name tbh, it feels to general while there might be other arithmethics not only this one that exposes `mul_by_generator` i would rather naem it AdditiveCurve?

crates/kdf/src/schema/additive.rs
```diff
@@ -0,0 +1,68 @@
...
+/// keys **do not** form a tree-like structure. Instead, child keys
+/// are all derived from a single root key and can be considered as
+/// "peers" to each other.
+#[autoimpl(Debug, Clone where C::PublicKey: trait)]
+#[derive(Copy)]
+pub struct Additive<C: Curve> {
+    master_pk: C::PublicKey,
+}
+
+impl<C> Additive<C>
```
pityjllk, line 19: AdditiveSchema?

crates/kdf/src/schema/borsh.rs
```diff
@@ -0,0 +1,99 @@
+use borsh::{BorshSerialize, io};
+#[cfg(feature = "digest")]
+pub use digest_io::IoWrapper;
```
pityjllk, line 3: use FQDNs

crates/kdf/src/schema/borsh.rs
```diff
@@ -0,0 +1,99 @@
...
+pub struct Borsh<W: WriteFinalizer = Vec<u8>>(W::Data);
+
+impl<W: WriteFinalizer> Borsh<W> {
+    #[inline]
+    pub const fn new(data: W::Data) -> Self {
+        Self(data)
+    }
+}
+
+impl<P, W> Schema<P> for Borsh<W>
```
pityjllk, line 26: do you see any usecases for 3rdparty implementations of WriteFinalizer?

it feels like overengineering a bit...


im wondering if much simpler 
```
pub struct Borsh<T = ()>(T);

impl<P> Schema<P> for Borsh<()>
where
    P: BorshSerialize,
{
...
}
```
and 
```
impl<P, D> Schema<P> for Borsh<D>
where
    P: BorshSerialize,
    D: digest::Digest + digest::Update + Clone,
{
...
}

```

wouldnt work as well

crates/kdf/src/schema/reduce.rs
```diff
@@ -0,0 +1,16 @@
+use core::marker::PhantomData;
+
+use impl_tools::autoimpl;
+
+/// [`Schema`](crate::Schema) for converting fixed byte arrays into a scalar
+/// via modular reduction.
+#[autoimpl(Debug, Clone, Default)]
+#[derive(Copy)]
+pub struct ReduceScalar<C>(PhantomData<C>);
```
pityjllk, line 8: doesnt reduce only applies to secp256 curve. maybe `ToScalar` ?

crates/kdf/crypto/src/ed25519.rs
```diff
@@ -0,0 +1,22 @@
...
+use crate::Curve;
+
+pub struct Ed25519;
+
+impl Curve for Ed25519 {
+    type PublicKey = VerifyingKey;
+
+    type Message = [u8];
+
+    type Signature = Signature;
```
pityjllk, line 12: reexport?

crates/kdf/crypto/src/secp256k1.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use crate::Curve;
+
+pub struct Secp256k1;
+
+impl Curve for Secp256k1 {
+    type PublicKey = VerifyingKey;
+    /// Prehash, i.e. output of a cryptographic hash function
+    type Message = [u8; 32];
+    type Signature = (Signature, RecoveryId);
```
pityjllk, line 14: reexport?

crates/outlayer/kdf-app/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+{
+    type Output = [u8; 32];
+
+    fn derive_path(&self, path: P) -> Self::Output {
+        const DOMAIN_SEPARATOR: &[u8] = b"outlayer/app-derivation/v1";
+
+        thread_local! {
+            // per-thread lazily-initialized hasher with pre-processed domain separator
+            static HASHER: Sha3_256 = Sha3_256::new_with_prefix(DOMAIN_SEPARATOR);
+        }
```
pityjllk, line 45: nit: honestly think its premature optimization... not worht the effort given the wasm runtime i think we are far from havig hashing function beehing our hot path

pityjllk (changes requested):

crates/kdf/src/schema/additive.rs
mitinarseny: I'd rather not have suffixes that indicate which traits this type implements...

crates/kdf/src/schema/borsh.rs
mitinarseny: This is a re-export

crates/kdf/src/schema/borsh.rs
mitinarseny: Unfortunately, Rust wouldn't allow you to have it due to "conflicting implementations for `Borsh<D>`", since we're not a `digest` crate itself, where these traits are defined

crates/kdf/src/schema/reduce.rs
mitinarseny: For ed25519, `Scalar::from_bytes_mod_order` also performs a reduction

crates/kdf/crypto/src/ed25519.rs
mitinarseny: Was thinking about that as well, but ended up with re-exporting the entire `ed25519_dalek` and `k256` if corresponding features are enabled. Wdyt?

crates/outlayer/kdf-app/src/lib.rs
mitinarseny: If WASM calls `sign` and/or `derive_public_key` many times in a cycle, then it can become so.
Anyway, why not to have it if it's already here?

crates/kdf/src/schema/additive.rs
mitinarseny: Basically, additive derivation is possible on all curves, where necessary arithmetic ops are implemented.
There is already the same abstraction [used](https://docs.rs/elliptic-curve/latest/elliptic_curve/trait.CurveArithmetic.html) in `k256` crate. This trait is a generalization of these from different curve crates. If we ever need more ops, we can add them here. Wdyt?

crates/kdf/src/schema/additive.rs
mitinarseny: `Additive` seems more important to me: `CurveArithmetic` is only needed to implement `Schema` for `Additive`

pityjllk: sure, just to give more context i find it  more readable when you can ready file top bottom (so later impl depends on primitives defined earlier)

crates/outlayer/kdf-app/src/lib.rs
mitinarseny: There is more: this crate can be used for offline derivation, where it would need to derive a lot of addresses per second

crates/kdf/src/schema/additive.rs
mitinarseny: It's interesting how people are used to process information in different ways! I personally prefer to first read some high-level definitions, and then, only if needed, dig into implementation details of it. Otherwise, my context window gets bloated with information that I might not needed if my intention was only to see those high-level definitions 😅

crates/outlayer/kdf-app/src/lib.rs
pityjllk: sure, its not that i wont be able sleep with that code on `main` branch :D

PR #276, opened 2026-05-25
chore(outlayer): read configs from env and introduce outlayer binary

pityjllk:
## Summary by CodeRabbit

* **New Features**
  * Added a worker binary and a configuration-driven service setup.
  * Exposed configurable executor, resolver, and cache settings (serde-backed).

* **Improvements**
  * Configurable memory and per-execution IO limits.
  * Optional per-execution fuel with a crate-level default.
  * HTTP resolver body-size limit is configurable.
  * Execution panics from blocking tasks are surfaced as an explicit error.

* **Refactor**
  * Builder APIs replaced with config-based builders for clearer construction and defaults.


crates/outlayer/service/src/config.rs
```diff
@@ -0,0 +1,27 @@
...
+#[cfg_attr(
+    feature = "serde",
+    derive(::serde::Serialize, ::serde::Deserialize),
+    serde(deny_unknown_fields, default)
+)]
+pub struct OutlayerConfig {
+    pub executor: ExecutorConfig,
+    pub resolver: ResolverConfig,
+    pub cache: CacheConfig,
+    pub default_fuel: u64,
```
hlgltvnnk, line 17: isn't it also an executor configuration parameter?

crates/outlayer/executor/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[cfg_attr(
+    feature = "serde",
+    derive(::serde::Serialize, ::serde::Deserialize),
+    serde(deny_unknown_fields, default)
+)]
+#[derive(Debug, Clone, Copy)]
+pub struct ExecutorConfig {
+    pub memory_limit: usize,
+    pub limits: ExecutorLimits,
```
hlgltvnnk: may be rename to StreamLimits?

crates/outlayer/executor/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub stdout: usize,
+    pub stderr: usize,
+}
+
+impl Default for ExecutorLimits {
+    fn default() -> Self {
+        Self {
+            stdin: 4 * 1024 * 1024,
+            stdout: 4 * 1024 * 1024,
+            stderr: 16 * 1024,
```
hlgltvnnk: may be make it as constants?

crates/outlayer/service/src/resolver/mod.rs
```diff
@@ -6,7 +6,26 @@ use bytes::Bytes;
 use defuse_outlayer_primitives::AppId;
 use sha2::{Digest, Sha256};
 
-use self::{near::NearResolver, url::UrlResolver};
+pub use self::near::NearResolver;
+pub use self::url::{HttpResolver, UrlResolver};
+
+#[cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(deny_unknown_fields, default))]
```
hlgltvnnk: #[cfg_attr(
    feature = "serde",
    derive(::serde::Serialize, ::serde::Deserialize),
    serde(deny_unknown_fields, default)
)]

crates/outlayer/service/src/resolver/mod.rs
```diff
@@ -6,7 +6,26 @@ use bytes::Bytes;
...
+    pub near_rpc_url: String,
+    pub near_chain_id: String,
+    pub http_max_len: usize,
+}
+
+impl Default for ResolverConfig {
+    fn default() -> Self {
+        Self {
+            near_rpc_url: "https://rpc.mainnet.near.org".to_owned(),
+            near_chain_id: "mainnet".to_owned(),
```
hlgltvnnk: may be do this with constants?

crates/outlayer/service/src/resolver/mod.rs
```diff
@@ -6,7 +6,26 @@ use bytes::Bytes;
...
 
-use self::{near::NearResolver, url::UrlResolver};
+pub use self::near::NearResolver;
+pub use self::url::{HttpResolver, UrlResolver};
+
+#[cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(deny_unknown_fields, default))]
+pub struct ResolverConfig {
+    pub near_rpc_url: String,
+    pub near_chain_id: String,
```
hlgltvnnk: may be use ChainId from near kit?

crates/outlayer/executor/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    pub fn build(self, signer: Arc<dyn Signer>) -> anyhow::Result<Executor> {
+        Ok(Executor::new(
+            signer,
+            Arc::new(VmRuntime::new(self.config.memory_limit)?),
+            self.config.limits,
+        ))
+    }
+}
```
hlgltvnnk, line 61: may be build it from config directly? like it is done with cache config?

Cargo.toml.orig
```diff
@@ -0,0 +1,195 @@
+[workspace]
+default-members = ["contracts/defuse"]
```
hlgltvnnk: pls this remove this file

crates/outlayer/service/src/lib.rs
```diff
@@ -64,20 +82,32 @@ impl Outlayer {
...
 
-        self.executor
-            .execute(
+        let executor = self.executor.clone();
+        let default_fuel = self.default_fuel;
+        // WASM execution is CPU-bound and wasmtime doesn't yield between instructions
+        // without epoch interruption, which would block the tokio scheduler.
+        // Run on the blocking thread pool to keep the async runtime responsive.
+        tokio::task::spawn_blocking(move || {
+            tokio::runtime::Handle::current().block_on(executor.execute(
```
hlgltvnnk: if executor.execute ever gets an await inside, this
blocking thread stays occupied for that whole time

hlgltvnnk (changes requested):

crates/outlayer/service/src/config.rs
pityjllk: the idea is to have optional fuel parameter that can be passed by a client. if its not passed then default value will be used. I dont think that default value for executor makes sense simply you call execute on low level `Executor` and you need to pass mandatory parameter `fuel`. but on higher level we want to make this parameter optional with some reasonable default thats why its configuration parameter of the Service and not Executor

crates/outlayer/service/src/resolver/mod.rs
pityjllk: not sure tbh, chain id is just string wrapper that does not give much benefits tbh. All in all it will be passed as a string in `.env` in our case either way. And it cant be wrapped easilly by `DisplayFromStr` because it does not implement `FromStr`.

https://github.com/r-near/near-kit-rs/pull/201

wdyt?

crates/outlayer/service/src/lib.rs
pityjllk: right, but at the moment we dont really have true async operations, just async interface. Also my understanding is that spawn_blocking spawns a thread dedicated for running cpu heavy sync operation (wasm execution). Sure we have async host api but its not really async now. Even if it would be it would just drive this async on this thread so it could not be reused BUT:
- there are 512 such threads in a pool by default
- there is no gain in running more than ~2*cpu cores wasm executions in parallell, as they are indeed cpu heavy. so thats far far far below the limit of 512. On the other hand running them in regular tokio task is far worse as cpu operations will starve the tokio scheduler

from what i see its possible to set some timestamp driven preemption of wasm execution, maybe thats the solution we might use once we have truly async host functions 🤔 but from all options considered this one feels okish, wdyt?

crates/outlayer/executor/src/config.rs
pityjllk: actully i did not like the pattern of calling build on config struct
- because i would expect `X::build` to return `X` not some other type
- in case of Outlayer the build parameter accepts signer, its even weirder when`X::build(foo)` so you passing `foo` to method build on `X` that returns some other type... so instead i added missing builder

```
    pub fn build(self, signer: impl Into<Arc<dyn Signer>>) -> anyhow::Result<Outlayer> {
        let executor = ExecutorBuilder::default()
            .with_config(self.config.executor)
            .build(signer.into())?;
        let resolver = ResolverBuilder::default()
            .with_config(self.config.resolver)
            .build();
        let cache = CacheBuilder::default()
            .with_config(self.config.cache)
            .build();
        Ok(Outlayer::new(
            resolver,
            executor,
            cache,
            self.config.default_fuel,
        ))

```

pityjllk: wdyt?

crates/outlayer/bin/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+        print_value(PREFIX, &defaults);
+        return Ok(());
+    }
+
+    dotenvy::dotenv().ok();
+    tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_env("RUST_LOG"))
+        .init();
+
+    let config: AppConfig = Config::builder()
```
hlgltvnnk: May be make it inside AppConfig implementation? And also accept file configuration
```
impl AppConfig {
    pub fn load() -> Result<Self, ConfigError> {
        let config_path = env::var("CONFIG_PATH").unwrap_or_else(|_| CONFIG_PATH.to_string());

        let settings = Config::builder()
            .add_source(
                File::with_name(&config_path)
                    .format(FileFormat::Toml)
                    .required(true),
            )
            .add_source(
                Environment::with_prefix("__")
                    .try_parsing(true)
                    .ignore_empty(true),
            )
            .build()?;

        settings.try_deserialize::<Self>()
    }
}
```

crates/outlayer/service/src/config.rs
```diff
@@ -0,0 +1,28 @@
+use defuse_outlayer_executor::ExecutorConfig;
+
+use crate::{CacheConfig, resolver::ResolverConfig};
```
hlgltvnnk: fuel is amount of instructions that will be executed - lets set it at least to 1_000_000_000 (as in old outlayer version)

crates/outlayer/bin/src/main.rs
```diff
@@ -0,0 +1,80 @@
...
+
+#[serde_with::serde_as]
+#[derive(Deserialize, Serialize, Default)]
+#[serde(default)]
+struct AppConfig {
+    #[serde(rename = "service")]
+    outlayer: OutlayerConfig,
+    #[serde_as(as = "Option<Hex>")]
+    signer_seed: Option<Vec<u8>>,
+}
```
hlgltvnnk: may be use SecretString from secrecy? or even zeroize?

crates/outlayer/service/src/cache.rs
```diff
@@ -3,46 +3,51 @@ use std::time::Duration;
...
-    max_capacity: u64,
-    time_to_idle: Option<Duration>,
+#[cfg_attr(
+    feature = "serde",
+    ::serde_with::serde_as,
+    derive(::serde::Serialize, ::serde::Deserialize),
+    serde(deny_unknown_fields, default)
+)]
+pub struct CacheConfig {
+    pub max_capacity: u64,
```
hlgltvnnk:
```suggestion
    pub max_capacity_bytes: u64,
```

crates/outlayer/service/src/lib.rs
hlgltvnnk: agree, currently that is ok (important to add concurrency limit on wasm execution), but if we will add true concurency to host functions it is worth tracking as a follow up before async host functions land - e.g. MPC signing over the network, the blocking thread won't be held waiting for I/O. 
So i think epoch interruption  makes WASM execution genuinely async: wasmtime returns `Pending` on each epoch tick, letting tokio schedule other tasks in between

hlgltvnnk (changes requested):

crates/outlayer/service/src/config.rs
hlgltvnnk: yes, it makes sense - in the future fuel should be calculated according to the fees in the request

crates/outlayer/service/src/resolver/mod.rs
pityjllk: ok, they merged and released new version so we can have ChainId now 🫡

hlgltvnnk: i think it should be used here because it provides defaults, so u dont need to specify it by yorself + also it [implements](https://github.com/r-near/near-kit-rs/blob/main/crates/near-kit/src/types/network.rs#L90) `FromStr` so it is compatible with DisplayFromStr

crates/outlayer/bin/src/main.rs
pityjllk: im not sure if we will need it. i gueess we will either run it from the test code where you can programically construct it or from docker compose that fetches configs from .env (phala cloud also requires docker compose)

crates/outlayer/bin/src/main.rs
pityjllk: went with zeroize

crates/outlayer/bin/src/main.rs
hlgltvnnk: it can be useful in local testing, but from another prspective u can just use .env - so lets just move current impl inside AppConfig so main will be more clean

crates/outlayer/service/src/lib.rs
hlgltvnnk: wdyt?

crates/outlayer/bin/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+        }
+        Value::Null => println!("# {prefix}="),
+        other => println!("{prefix}={other}"),
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    //TODO: is that even needed?
+    if std::env::args().any(|a| a == "--print-env") {
```
hlgltvnnk: Is that done for debuging purposes or just as an example? May be just add .env.example?

crates/outlayer/service/src/lib.rs
pityjllk: Even if we would have this epoch interuptions (fine tuning correct period is not trivial either way) it still doesnt feel like a good idea to run it in regular tokio task 

Lets say we use this epoch interuptions. its still cpu heavy task, and no matter if its 1 call that takes 50ms of 50 concecutive for 1ms it can still starve the scheduler. My understanding is that tokio wins where there is a lot of waiting/async operations and the rest that happens between this "waiting" are trivial operations. In our case is legitimatie CPU heavy wasm execution and for this reason i think it should be run in spawn_blocking either way. There are still some advantages of interruptions (timeouting execution etc). 

But for the reasons state above i would rather keep it as it is. wdyt

crates/outlayer/bin/src/main.rs
pityjllk: i decided to keep it for some time unit all configuration parameters (and their names) settles 🤔 wdyt?

crates/outlayer/service/src/lib.rs
hlgltvnnk: agree, with that approach wasm execution will cause starvation

crates/outlayer/service/src/cache.rs
```diff
@@ -3,46 +3,51 @@ use std::time::Duration;
...
+    )]
+    pub time_to_idle: Option<Duration>,
 }
 
-impl Default for CacheBuilder {
+impl Default for CacheConfig {
     fn default() -> Self {
         Self {
-            max_capacity: DEFAULT_MAX_CAPACITY,
+            max_capacity_bytes: 100 * 1024 * 1024,
```
hlgltvnnk: make it constant please

hlgltvnnk (dismissed):

crates/outlayer/bin/src/main.rs
hlgltvnnk: may be i just dont understand the purpose of it, if u want to check default values - u can check it in code and if u need some example - it should be in .env.example

Cargo.toml
```diff
@@ -157,6 +160,7 @@ near-workspaces = "0.22"
 anyhow = "1"
 arbitrary = { version = "1", features = ["derive"] }
 array-util = "1"
+async-nats = "0.46.0"
```
mitinarseny: Is it needed anymore?

crates/outlayer/bin/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "defuse-outlayer-bin"
```
mitinarseny: Maybe `worker`? There can be many outlayer-related binaries in the future
```suggestion
name = "defuse-outlayer-worker"
```

crates/outlayer/worker/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+impl AppConfig {
+    fn load() -> Result<Self> {
+        Config::builder()
+            .add_source(
+                Environment::with_prefix(PREFIX)
+                    .prefix_separator("__")
+                    .separator("__"),
+            )
```
mitinarseny, line 31: Can/should we add `worker.toml` config source as well?

crates/outlayer/worker/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+        None => unimplemented!("signer seed must be provided until CKD integration is complete"),
+    };
+
+    let signer: Arc<dyn Signer> = Arc::new(signer);
+    let _ = OutlayerBuilder::default()
+        .with_config(outlayer)
+        .build(signer)
+        .context("outlayer")?;
+
+    Ok(())
```
mitinarseny: Something like `service.run().await?`?

crates/outlayer/executor/src/config.rs
```diff
@@ -0,0 +1,71 @@
...
+
+impl Default for StreamLimits {
+    fn default() -> Self {
+        Self {
+            stdin: LIMIT_4MB,
+            stdout: LIMIT_4MB,
+            stderr: LIMIT_16KB,
+        }
+    }
+}
```
mitinarseny, line 51: Maybe to have defaults on each field as well, so that, for example, if only `stdin` is specified, then other two get the default value...?

Also, I'd suggest all memory-limit-related fields to have `_bytes` postfix, so it's less likely to misconfigure or misuse the value

crates/outlayer/executor/src/config.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[cfg_attr(
+    feature = "serde",
+    derive(::serde::Serialize, ::serde::Deserialize),
+    serde(deny_unknown_fields, default)
+)]
+#[derive(Debug, Clone, Copy)]
+pub struct ExecutorConfig {
+    pub memory_limit: usize,
+    pub limits: StreamLimits,
```
mitinarseny:
```suggestion
    pub io_limits: IoLimits,
```

crates/outlayer/executor/src/lib.rs
```diff
@@ -20,6 +21,7 @@ use bytes::Bytes;
 pub struct Executor {
     runtime: Arc<VmRuntime>,
     signer: Arc<dyn Signer>,
+    limits: StreamLimits,
```
mitinarseny:
```suggestion
    io_limits: IoLimits,
```

crates/outlayer/executor/src/config.rs
```diff
@@ -0,0 +1,71 @@
...
+const LIMIT_16KB: usize = 16 * 1024;
+
+#[cfg_attr(
+    feature = "serde",
+    derive(::serde::Serialize, ::serde::Deserialize),
+    serde(deny_unknown_fields, default)
+)]
+#[derive(Debug, Clone, Copy)]
+pub struct ExecutorConfig {
+    pub memory_limit: usize,
```
mitinarseny:
```suggestion
    pub memory_limit_bytes: usize,
```

crates/outlayer/executor/src/config.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    pub fn build(self, signer: Arc<dyn Signer>) -> anyhow::Result<Executor> {
+        Ok(Executor::new(
+            signer,
+            Arc::new(VmRuntime::new(self.config.memory_limit)?),
+            self.config.limits,
+        ))
+    }
+}
```
mitinarseny: Is it really needed? You can have `ExecutorConfig::build(signer)` doing the same instead

crates/outlayer/worker/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+const PREFIX: &str = "WORKER";
+
+#[serde_with::serde_as]
+#[derive(Deserialize, Default)]
+#[serde(default)]
+struct AppConfig {
+    #[serde(rename = "service")]
+    outlayer: OutlayerConfig,
+    #[serde_as(as = "Option<Hex>")]
+    signer_seed: Option<Zeroizing<Vec<u8>>>,
```
mitinarseny: `seed` can be used not only for signer, but for encrypt/decrypt/KDF, etc..
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    #[serde_as(as = "Option<Hex>")]
    seed: Option<Zeroizing<Vec<u8>>>,
```

crates/outlayer/worker/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+defuse-outlayer-host = { workspace = true }
+defuse-outlayer-service = { workspace = true }
+defuse-outlayer-signer = { workspace = true, features = ["ed25519", "secp256k1"] }
+
+anyhow.workspace = true
+config.workspace = true
+dotenvy.workspace = true
+serde = { workspace = true, features = ["derive"] }
+serde_with = { workspace = true, features = ["hex"] }
+tokio = { workspace = true, features = ["full"] }
```
mitinarseny: Let's enable features only when needed
```suggestion
tokio = { workspace = true, features = ["macros", "rt"] }
```

crates/outlayer/worker/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+            .context("config")
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    dotenvy::dotenv().ok();
+    tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_env("RUST_LOG"))
+        .init();
```
mitinarseny, line 43:
```suggestion
    tracing_subscriber::fmt()
        .with_env_filter(tracing_subscriber::EnvFilter::from_env("OUTLAYER_LOG"))
        .init();
```

In one of the next PRs, let's add:
* override for `OUTLAYER_LOG` via CLI args
* [OpenTelemetry support](https://crates.io/crates/tracing-opentelemetry)
* Tee'ing logs to stderr as JSON/text based on the [`is_terminal()`](https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html#tymethod.is_terminal)

crates/outlayer/service/src/resolver/url/http.rs
```diff
@@ -16,6 +16,13 @@ pub struct HttpResolver {
 }
 
 impl HttpResolver {
+    pub fn new(max_len: usize) -> Self {
```
mitinarseny:
```suggestion
    pub fn new(max_len_bytes: usize) -> Self {
```

crates/outlayer/executor/Cargo.toml
```diff
@@ -19,6 +19,9 @@ cfg_eval = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
 serde_with = { workspace = true, features = ["base64"], optional = true }
 
+[dev-dependencies]
+serde_json.workspace = true
```
mitinarseny: Hmm, I don't see any new tests added in this PR, is it still needed?

crates/outlayer/service/src/resolver/config.rs
```diff
@@ -0,0 +1,46 @@
...
+    ::cfg_eval::cfg_eval,
+    ::serde_with::serde_as,
+    derive(::serde::Serialize, ::serde::Deserialize),
+    serde(deny_unknown_fields, default)
+)]
+pub struct ResolverConfig {
+    pub near_rpc_url: String,
+    #[cfg_attr(feature = "serde", serde_as(as = "::serde_with::DisplayFromStr"))]
+    pub near_chain_id: near_kit::ChainId,
+    pub http_max_len: usize,
```
mitinarseny: 1. Can we please add `_bytes` postfix?
2. It's not clear, what does "HTTP len" mean: are response headers included in this limit?
```suggestion
    pub http_max_len_bytes: usize,
```

crates/outlayer/service/src/resolver/config.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[cfg_attr(
+    feature = "serde",
+    ::cfg_eval::cfg_eval,
+    ::serde_with::serde_as,
+    derive(::serde::Serialize, ::serde::Deserialize),
+    serde(deny_unknown_fields, default)
+)]
+pub struct ResolverConfig {
+    pub near_rpc_url: String,
```
mitinarseny:
```suggestion
    #[serde(default = "mainnet_rpc_url")]
    pub near_rpc_url: Url,
    
// use this function in both: `#[serde(default)]` and `impl Default for ResolverConfig`
fn mainnet_rpc_url() -> Url {
    // return parsed hardcoded mainnet rpc url here
}
```

crates/outlayer/service/src/resolver/config.rs
```diff
@@ -0,0 +1,46 @@
...
+    #[cfg_attr(feature = "serde", serde_as(as = "::serde_with::DisplayFromStr"))]
+    pub near_chain_id: near_kit::ChainId,
```
mitinarseny: `near_kit::ChainId` is just a wrapper around the string. Using `near-kit` crate is an implementation detail, so better not expose this type in the config
```suggestion
    pub near_chain_id: String,
```

crates/outlayer/service/src/resolver/config.rs
```diff
@@ -0,0 +1,46 @@
...
+        self
+    }
+
+    pub fn build(self) -> Resolver {
+        let near = near_kit::Near::custom(self.0.near_rpc_url, self.0.near_chain_id).build();
+        let near = NearResolver::new(near);
+        let url = UrlResolver::new(HttpResolver::new(self.0.http_max_len));
+        Resolver::new(near, url)
+    }
+}
```
mitinarseny: Seems redundant, the same can be done in `ResolverConfig::build()`.

crates/outlayer/service/src/resolver/mod.rs
```diff
@@ -6,7 +7,9 @@ use bytes::Bytes;
 use defuse_outlayer_primitives::AppId;
 use sha2::{Digest, Sha256};
 
-use self::{near::NearResolver, url::UrlResolver};
+pub use self::config::{ResolverBuilder, ResolverConfig};
+pub use self::near::NearResolver;
+pub use self::url::{HttpResolver, UrlResolver};
```
mitinarseny: merge pls

crates/outlayer/service/src/builder.rs
```diff
@@ -1,24 +1,36 @@
...
+            .with_config(self.config.cache)
+            .build();
+        Ok(Outlayer::new(
             resolver,
             executor,
-            runtime_cache: self.cache.build(),
-        }
+            cache,
+            self.config.default_fuel,
+        ))
```
mitinarseny: All these builders seem redundant, because they only store the `config`, and then build based on this config. It can be methods on corresponding Config types

crates/outlayer/service/src/cache.rs
```diff
@@ -3,46 +3,53 @@ use std::time::Duration;
...
+    ::serde_with::serde_as,
+    derive(::serde::Serialize, ::serde::Deserialize),
+    serde(deny_unknown_fields, default)
+)]
+pub struct CacheConfig {
+    pub max_capacity_bytes: u64,
+    #[cfg_attr(
+        feature = "serde",
+        serde_as(as = "Option<::serde_with::DurationSeconds<u64>>")
+    )]
```
mitinarseny, line 19:
```suggestion
    #[cfg_attr(
        feature = "serde",
        serde(rename = "time_to_idle_seconds", default, skip_serializing_if = "Option::is_none"),
        serde_as(as = "Option<::serde_with::DurationSeconds<u64>>")
    )]
```

crates/outlayer/service/src/cache.rs
```diff
@@ -3,46 +3,53 @@ use std::time::Duration;
...
-    max_capacity: u64,
-    time_to_idle: Option<Duration>,
+#[cfg_attr(
+    feature = "serde",
+    ::serde_with::serde_as,
+    derive(::serde::Serialize, ::serde::Deserialize),
+    serde(deny_unknown_fields, default)
+)]
+pub struct CacheConfig {
+    pub max_capacity_bytes: u64,
```
mitinarseny, line 14:
```suggestion
    #[serde(default = "default_capacity_bytes")]
    pub max_capacity_bytes: u64,
```

crates/outlayer/service/src/cache.rs
```diff
@@ -3,46 +3,53 @@ use std::time::Duration;
...
     }
 }
 
-impl CacheBuilder {
-    pub const fn max_capacity(mut self, max_capacity: u64) -> Self {
-        self.max_capacity = max_capacity;
-        self
-    }
+#[derive(Default)]
+pub struct CacheBuilder(CacheConfig);
```
mitinarseny: redundant

crates/outlayer/service/src/config.rs
```diff
@@ -0,0 +1,28 @@
...
+#[cfg_attr(
+    feature = "serde",
+    derive(::serde::Serialize, ::serde::Deserialize),
+    serde(deny_unknown_fields, default)
+)]
+pub struct OutlayerConfig {
+    pub executor: ExecutorConfig,
+    pub resolver: ResolverConfig,
+    pub cache: CacheConfig,
+    pub default_fuel: u64,
```
mitinarseny, line 17:
```suggestion
    #[serde(default = "default_fuel", skip_serializing_if = "Option::is_none")]
    pub default_fuel: Option<u64>,
```

crates/outlayer/service/src/lib.rs
```diff
@@ -21,11 +26,22 @@ pub struct Outlayer {
...
+    default_fuel: u64,
 }
 
 impl Outlayer {
-    pub fn builder() -> OutlayerBuilder {
-        OutlayerBuilder::default()
+    pub const fn new(
+        resolver: Resolver,
+        executor: Executor,
+        runtime_cache: Cache<[u8; 32], Component>,
```
mitinarseny: It's leaking the abstraction: cache should be initialized purely within the constructor, from some configuration params passed to this constructor

Cargo.toml
```diff
@@ -133,6 +135,7 @@ defuse-outlayer-sdk.path = "crates/outlayer/sdk"
 defuse-outlayer-service.path = "crates/outlayer/service"
 defuse-outlayer-signer.path = "crates/outlayer/signer"
 defuse-outlayer-sys.path = "crates/outlayer/sys"
+defuse-outlayer-utils.path = "crates/outlayer/utils"
```
mitinarseny: What is this?)

Cargo.toml
```diff
@@ -167,11 +171,13 @@ bytes = "1.11.1"
 cfg_eval = "0.1.2"
 chrono = { version = "0.4", default-features = false }
 clap = { version = "4.5.55", features = ["derive"] }
+config = { version = "0.15", default-features = false, features = ["json"] }
```
mitinarseny: Activate JSON feature only in the crate that needs it
```suggestion
config = { version = "0.15", default-features = false }
```

crates/outlayer/service/src/lib.rs
mitinarseny: In this case, can async host-functions run on threads allocated by tokio for non-blocking tasks?

mitinarseny (changes requested):

crates/outlayer/worker/src/main.rs
pityjllk: what would be usecase for it? I think we will either set up some rust tests where service will be created programically or we will spin up some e2e tests most likely in docker-compose 🤔 we might add it in a future (file based config is supported) when we will find legitimate usecase for it, but for now it doesnt seem needed

crates/outlayer/worker/src/main.rs
pityjllk: added in follow up pr

crates/outlayer/executor/src/config.rs
pityjllk: programically you can simply
```
impl Default for StreamLimits {
    fn default() -> Self {
        Self {
            stdin: LIMIT_4MB,
, ..Default::default()
        }
    }
}
```

while the config crate start with Default values and overrides only passed ones

crates/outlayer/executor/src/config.rs
pityjllk: removed here and in other places

crates/outlayer/worker/src/main.rs
pityjllk: renamed, but these config structs only need deserialization s o`skip_serializing_if` is redundant

crates/outlayer/worker/Cargo.toml
pityjllk: set to "macros", "rt-multi-threaded"

crates/outlayer/executor/Cargo.toml
pityjllk: right, sorry some leftover from some local benchmarks i was doing

crates/outlayer/service/src/resolver/config.rs
pityjllk: renamed to `max_body_bytes`

Headers don't need to be counted here because hyper bounds the response head during parsing, so oversized headers fail before we ever get a `Response`:
- ≤ 100 header fields — [`DEFAULT_MAX_HEADERS`](https://github.com/hyperium/hyper/blob/v1.9.0/src/proto/h1/role.rs#L31)
- ≤ ~408 KiB total head buffer (`8192 + 4096 * 100`) — [`DEFAULT_MAX_BUFFER_SIZE`](https://github.com/hyperium/hyper/blob/v1.9.0/src/proto/h1/io.rs#L23); the parser returns `TooLarge` once the read buffer exceeds it.

pityjllk: and at the time we start counting headers were alredy received

crates/outlayer/service/src/resolver/config.rs
pityjllk: (changed to url)


whats the advantaqge of that? There alaready is. impl Default for config strcut, and struct is annotated wiht #[serde(default)] so it simply picks up default value from `impl Default` right?

crates/outlayer/executor/src/config.rs
mitinarseny: yeah but what if need to deserialize these configs from files later and opt-out from `config`?

crates/outlayer/worker/src/main.rs
mitinarseny: What about `outlayer-worker print-default-cfg`?

crates/outlayer/service/src/resolver/config.rs
mitinarseny: Unfortunately, it doesn't. That's why I'm suggesting it all with `#[serde(default = "some_default")]`

crates/outlayer/executor/src/config.rs
pityjllk: i guess then we can introduce it ? wdyt?

crates/outlayer/worker/src/main.rs
pityjllk: so i used to have this `print-env` option but i removed it actually, as i concluded that example.env should be enough 🤔 wdyt?

crates/outlayer/service/src/resolver/config.rs
pityjllk:
```

mod tests {
    use super::*;

    #[test]
    fn parse_config() {
        // parse ResolverConfig from empty json
        let c = serde_json::from_str::<ResolverConfig>("{}").unwrap();
        assert_eq!(c.near_rpc_url.as_str(), NEAR_RPC_URL);
        assert_eq!(c.near_chain_id.as_str(), NEAR_CHAIN_ID);
        assert_eq!(c.http_max_body_bytes, MAX_WASM_SIZE_10MB);
    }
}

```
this works just fine, or am i missing something?

mitinarseny: yeah but what if you have only 1 of these 3 values set?

crates/outlayer/worker/src/main.rs
mitinarseny: I'm not suggesting to have this `print-cfg/env` command right now, but:
AFAIU, env variables are the only way to set config params so that it doesn't change TEE measurements (correct me if I'm wrong). If so, then we would want to have some config params set that cannot be changed (e.g. whitelist of outlayer-app contract instances), while others can be (e.g. ??)

crates/outlayer/service/src/resolver/config.rs
pityjllk:
```
mod tests {
    use super::*;

    #[test]
    fn parse_config() {
        // parse ResolverConfig from empty json
        let c = serde_json::from_str::<ResolverConfig>("{\"near_rpc_url\": \"https://example.com/\"}").unwrap();
        assert_eq!(c.near_rpc_url.as_str(), "https://example.com/");
        assert_eq!(c.near_chain_id.as_str(), NEAR_CHAIN_ID);
        assert_eq!(c.http_max_body_bytes, MAX_WASM_SIZE_10MB);
    }
}

```

still seems to work just fine

crates/outlayer/executor/src/config.rs
mitinarseny: The current impl of `Deserialize` is misleading. So if we have it now already, I suggest to make it behave as intended from the beginning rather than having surprises later

crates/outlayer/service/src/resolver/config.rs
mitinarseny: Hmmm... Is it because of `#[serde(default)]` on the struct level?

crates/outlayer/worker/src/main.rs
pityjllk: agreed that we will go with env and deserialize only approach for now

crates/outlayer/executor/src/config.rs
pityjllk: agreed that we will go with env and deserialize only approach for now

crates/outlayer/worker/src/main.rs
pityjllk: >env variables are the only way to set config params so that it doesn't change TEE measurements

if its defined in docker-compose explicitally they are part of attestation, you may opt out from it if you configure them as secrets, the sad part is that if you want to update secrets you need to always update all of them (not only one variable)

pityjllk: agreed that we will go with env and deserialize only approach for now

crates/outlayer/service/src/cache.rs
pityjllk: agreed that we will go with env and deserialize only approach for now

crates/outlayer/service/src/cache.rs
pityjllk: agreed that we will go with env and deserialize only approach for now

crates/outlayer/service/src/config.rs
pityjllk: agreed that we will go with env and deserialize only approach for now

crates/outlayer/executor/src/lib.rs
```diff
@@ -50,12 +52,15 @@ impl Outcome {
     }
 }
 
-// TODO: maybe read from config?
-const STDIN_LIMIT: usize = 4 * 1024 * 1024; // 4 MB
-const STDOUT_LIMIT: usize = 4 * 1024 * 1024; // 4 MB
-const STDERR_LIMIT: usize = 16 * 1024; // 16 KB
-
 impl Executor {
+    pub fn new(signer: Arc<dyn Signer>, runtime: Arc<VmRuntime>, io_limits: IoLimits) -> Self {
```
mitinarseny: Why does `clippy` do not issue a warning for it?
```suggestion
    pub const fn new(signer: Arc<dyn Signer>, runtime: Arc<VmRuntime>, io_limits: IoLimits) -> Self {
```

crates/outlayer/executor/src/config.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct ExecutorConfig {
+    pub memory_limit_bytes: usize,
+    pub io_limits: IoLimits,
+}
+
+impl Default for ExecutorConfig {
+    fn default() -> Self {
+        Self {
+            memory_limit_bytes: VmRuntime::DEFAULT_MEMORY_LIMIT,
```
mitinarseny: So, it's not a "default memory limit" of `VmRuntime` anymore - it's only used by `ExecutorConfig`. Maybe then move `DEFAULT_MEMORY_LIMIT` here as well?

crates/outlayer/service/src/cache.rs
```diff
@@ -3,38 +3,37 @@ use std::time::Duration;
...
-#[must_use = "use .build()"]
-#[derive(Debug, Clone)]
-pub struct CacheBuilder {
-    max_capacity: u64,
-    time_to_idle: Option<Duration>,
+#[cfg_attr(
+    feature = "serde",
+    ::serde_with::serde_as,
+    derive(::serde::Deserialize),
+    serde(deny_unknown_fields, default)
```
mitinarseny, line 11: `cfg_eval`?

crates/outlayer/service/src/cache.rs
```diff
@@ -3,38 +3,37 @@ use std::time::Duration;
...
 }
 
-impl Default for CacheBuilder {
+const DEFAULT_MAX_CAPACITY_BYTES: u64 = 100 * 1024 * 1024;
+
+impl Default for CacheConfig {
     fn default() -> Self {
         Self {
-            max_capacity: DEFAULT_MAX_CAPACITY,
+            max_capacity_bytes: DEFAULT_MAX_CAPACITY_BYTES,
```
mitinarseny: nit: rename or have a comment `100 MiB`

crates/outlayer/worker/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+[lints]
+workspace = true
+
+[[bin]]
+name = "outlayer-worker"
+path = "src/main.rs"
+
+[dependencies]
+defuse-outlayer-host = { workspace = true }
+defuse-outlayer-service = { workspace = true }
```
mitinarseny: Maybe re-export `host` from `service`?

crates/outlayer/service/src/lib.rs
```diff
@@ -64,20 +76,32 @@ impl Outlayer {
...
+        app: Code<'_>,
+        input: Bytes,
+        fuel: Option<u64>,
+    ) -> Result<Outcome, Error> {
         let (app_id, component) = self.resolve_app(app).await?;
 
-        self.executor
-            .execute(
+        let executor = self.executor.clone();
+        let default_fuel = self.default_fuel;
```
mitinarseny: unwrap it right here:
```rust
let fuel = fuel.unwrap_or(self.default_fuel)
```

crates/outlayer/worker/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+use defuse_outlayer_service::OutlayerConfig;
+use defuse_outlayer_signer::InMemorySigner;
+
+use anyhow::{Context as _, Result};
+use config::{Config, Environment};
+use serde::Deserialize;
+use serde_with::hex::Hex;
+use zeroize::Zeroizing;
+
+const PREFIX: &str = "WORKER";
```
mitinarseny: For consistency with `OUTLAYER_LOG`:
```suggestion
const PREFIX: &str = "OUTLAYER";
```

crates/outlayer/worker/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    dotenvy::dotenv().ok();
+    tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_env("OUTLAYER_LOG"))
+        .init();
+
+    let AppConfig { outlayer, seed } = AppConfig::load()?;
```
mitinarseny:
```suggestion
    let AppConfig { outlayer, seed } = AppConfig::load().context("config")?;
```

crates/outlayer/worker/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    let signer = match seed.as_deref() {
+        Some(seed) => {
+            tracing::warn!("using custom signer seed — not intended for production use");
+            InMemorySigner::from_seed(seed)
+        }
+        None => unimplemented!("signer seed must be provided until CKD integration is complete"),
+    };
+
+    let signer: Arc<dyn Signer> = Arc::new(signer);
+    let _ = outlayer.build(signer).context("outlayer")?;
```
mitinarseny:
```suggestion
    let _ = outlayer.build(signer).context("build")?;
```

crates/outlayer/executor/src/lib.rs
pityjllk: it seems like due to trait object or ita s bug in clippy

crates/outlayer/worker/Cargo.toml
pityjllk: ok

crates/outlayer/worker/src/main.rs
pityjllk: fyi context config is already added in  AppConfig::load

mitinarseny: IMO it's a wrong place to have it, context should be added by callers, not by callees.

crates/outlayer/service/Cargo.toml
```diff
@@ -10,6 +10,7 @@ workspace = true
 
 [dependencies]
 defuse-outlayer-executor.workspace = true
+defuse-outlayer-host.workspace = true
```
mitinarseny: Oh, I mean, re-export from executor, and then - from service...? Like, where `dyn Signer` is actually used

crates/outlayer/service/src/lib.rs
```diff
@@ -96,7 +98,7 @@ impl Outlayer {
                     host: HostContext { app_id },
                 },
                 &component,
-                fuel.unwrap_or(default_fuel),
+                fuel,
             ))
```
mitinarseny: nit:
```rust
tokio::task::spawn_blocking({
    // clone the all things here, so that it clearly indicates you're doing so because of `move` below
    let executor = self.executor.clone();
    move || tokio::runtime::Handle::current().block_on(...).await
})
```
And also, maybe `let ctx = Context {...}` there, before the `move`, as well?

mitinarseny (approved):

crates/outlayer/service/src/lib.rs
pityjllk: it seems that asyncs are always run on single scheduler either way 👍

Cargo.toml
```diff
@@ -52,6 +52,7 @@ members = [
   "crates/testing/randomness",
   "crates/testing/sandbox",
 
+  "crates/outlayer/worker",
```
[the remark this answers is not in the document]
pityjllk: removed

PR #277, opened 2026-05-27
feat: grpc server

pityjllk:
reflections

```
> grpcurl -plaintext localhost:50051 list

grpc.health.v1.Health
grpc.reflection.v1.ServerReflection
outlayer.OutlayerService
```

checkhealth for seamless integration with kubernetes/LB
```
> grpcurl -plaintext -d '{"service":""}' localhost:50051 grpc.health.v1.Health/Check

{
  "status": "SERVING"

```

wasm request execution
```
> grpcurl -plaintext \
        -d '{"app": { "code_ref": { "app_id": { "near": "0s1fe35ffee209f01f3a508c933d2efec883dfd083" }}}}' \
        localhost:50051 \
        outlayer.OutlayerService/Execute
{
  "execution": {
    "details": {
      "fuelConsumed": "1246"
    }
  }
}

```

## Summary by CodeRabbit

* **New Features**
  * Added a gRPC interface and request/response messaging for the Outlayer worker, with health and reflection support.
  * Built-in server protections: load shedding, buffering, concurrency limits, and timeouts.
  * Configurable gRPC server options (address, per-connection concurrency) and observability via tracing on executions.

* **Chores**
  * CI and docs workflows now install protobuf tooling required for building and publishing proto-based artifacts.


crates/outlayer/bin/src/main.rs
```diff
@@ -1,25 +1,58 @@
...
 #[serde(default)]
 struct AppConfig {
     #[serde(rename = "service")]
     outlayer: OutlayerConfig,
     #[serde_as(as = "Option<Hex>")]
     signer_seed: Option<Zeroizing<Vec<u8>>>,
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    addr: SocketAddr,
+    concurrency_limit: usize,
+    connections_limit: usize,
```
mitinarseny: The difference between these two is not clear from the naming tbh

crates/outlayer/worker/src/main.rs
```diff
@@ -1,25 +1,58 @@
...
     outlayer: OutlayerConfig,
     #[serde_as(as = "Option<Hex>")]
     signer_seed: Option<Zeroizing<Vec<u8>>>,
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    addr: SocketAddr,
+    concurrency_limit: usize,
+    connections_limit: usize,
+    concurrency_limit_per_connection: usize,
+    #[serde_as(as = "serde_with::DurationSeconds<u64>")]
+    execution_timeout_s: Duration,
```
mitinarseny:
```suggestion
    #[serde_as(rename = "execution_timeout_seconds", as = "serde_with::DurationSeconds<u64>")]
    execution_timeout: Duration,
```

mitinarseny: What is it supposed to mean btw? How long a WASM itself is executed? Do we take into account time for resolving?

crates/outlayer/proto/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+#![allow(clippy::all, clippy::pedantic, clippy::nursery)]
+
+tonic::include_proto!("outlayer");
+
+pub const FILE_DESCRIPTOR_SET: &[u8] = tonic::include_file_descriptor_set!("outlayer_descriptor");
```
mitinarseny, line 12: can you pls add a comment on what this is for?

crates/outlayer/proto/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+edition.workspace = true
+version.workspace = true
+rust-version.workspace = true
+repository.workspace = true
+publish = false
+
+[dependencies]
+prost = { workspace = true }
+tonic = { workspace = true }
+tonic-prost = { workspace = true }
```
mitinarseny, line 15: I can see only `tonic::` used in `src/lib.rs`, no? If they're still needed, pls add a comment describing why

crates/outlayer/proto/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+
+[build-dependencies]
+tonic-prost-build = "0.14"
```
mitinarseny: Maybe add to workspace deps?

crates/outlayer/vm-runner/Cargo.toml
```diff
@@ -11,10 +11,11 @@ workspace = true
 [dependencies]
 defuse-outlayer-host.workspace = true
 
+defuse-outlayer-proto = { workspace = true, optional = true }
```
mitinarseny: Can we avoid poisoning all dependent crates with it? Can't we just convert them at the level of gRPC server?
Imagine we have HTTP server, too. Would we add conversions from/to OpenAPI spec here, too?

crates/outlayer/service/Cargo.toml
```diff
@@ -31,10 +33,26 @@ url.workspace = true
...
+  "dep:defuse-outlayer-proto",
+]
+tonic = [
+  "dep:tonic",
+  "proto",
+  "tower",
+  "tower/load-shed",
+  "tower/timeout",
+  "tower/util",
+]
```
mitinarseny: It's a bit weird to have this feature... Maybe to have a separate crate for gRPC service?

crates/outlayer/service/src/tonic_impl.rs
```diff
@@ -0,0 +1,53 @@
...
+            } else {
+                Status::internal(e.to_string())
+            }
+        })?;
+
+        Ok(Response::new(proto::Response {
+            kind: Some(proto::response::Kind::Execute(outcome.into())),
+        }))
+    }
+}
```
mitinarseny: So, this service only converts proto request definitions to underlying types expected by the inner service and then converts the response back to proto? If so, then why does it need to be generic over `S`?
I guess, to be able to use load_shed/timeout underneath, etc... But can we just apply all these load_shed/timeouts **on top** of this service, not between this one and the actual `Executor`

crates/outlayer/bin/src/main.rs
pityjllk: renamed concurency_limit to max_parallel_wasm_executions

crates/outlayer/worker/src/main.rs
pityjllk: renamed to 'reqeust_handling_timeout`, its overall timeout to at least have someway to unstuck the worker in corner case.

crates/outlayer/proto/Cargo.toml
pityjllk: added comment

crates/outlayer/grpc/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub fn new(outlayer: Outlayer, config: GrpcConfig) -> Self {
+        let service = ServiceBuilder::new()
+            // When the queue is full, immediately return RESOURCE_EXHAUSTED so the load
+            // balancer can route the request elsewhere rather than waiting here.
+            // RESOURCE_EXHAUSTED is the standard gRPC backpressure code respected by
+            // gRPC-aware load balancers (e.g. Envoy, AWS ALB).
+            .load_shed()
+            // Bounded async queue. Decouples acceptance from execution. When full, the
+            // load_shed layer above fires.
+            .buffer(config.connections_limit)
```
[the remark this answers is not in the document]
pityjllk: if it fails it fails

crates/outlayer/vm-runner/Cargo.toml
pityjllk: reworked so all proto impls lives in grpc now

crates/outlayer/service/src/tonic_impl.rs
pityjllk: changed, so that its grpc service that decorates outlayer service now

crates/outlayer/proto/proto/outlayer.proto
```diff
@@ -0,0 +1,68 @@
+syntax = "proto3";
+package outlayer;
```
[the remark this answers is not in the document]
pityjllk: fp we dont use it yet

crates/outlayer/service/src/tower_impl.rs
```diff
@@ -0,0 +1,31 @@
...
+pub struct ExecuteRequest {
+    pub app: Code<'static>,
+    pub input: Bytes,
+    pub fuel: Option<u64>,
+}
+
+impl tower::Service<ExecuteRequest> for Outlayer {
+    type Response = Outcome;
+    type Error = Error;
+    type Future = Pin<Box<dyn Future<Output = Result<Outcome, Error>> + Send>>;
```
mitinarseny: Use `futures` crate:
```suggestion
    type Future = BoxFuture<'static, Result<Outcome, Error>>;
```

crates/outlayer/service/src/tower_impl.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: ExecuteRequest) -> Self::Future {
+        let this = self.clone();
+        Box::pin(async move { this.execute(req.app, req.input, req.fuel).await })
```
mitinarseny: Use `FutureExt::boxed()` instead of doing it manually
Also, maybe `.instrument()` it with the current span?

crates/outlayer/service/src/tower_impl.rs
```diff
@@ -0,0 +1,31 @@
...
+use crate::{Code, Error, Outlayer};
+use defuse_outlayer_executor::Outcome;
+
+pub struct ExecuteRequest {
+    pub app: Code<'static>,
+    pub input: Bytes,
+    pub fuel: Option<u64>,
+}
+
+impl tower::Service<ExecuteRequest> for Outlayer {
```
mitinarseny: What's the purpose of this implementation if it could be inside gRPC service impl itself?

crates/outlayer/grpc/src/convert.rs
```diff
@@ -0,0 +1,121 @@
...
+use defuse_outlayer_service::{AppCodeUrl, Code, CodeRef, ExecuteRequest};
+use defuse_outlayer_vm_runner::{ExecutionDetails, ExecutionOutcome};
+
+pub trait TryFromProto<P>: Sized {
+    fn try_from_proto(proto: P) -> anyhow::Result<Self>;
+}
+
+pub trait IntoProto<P> {
+    fn into_proto(self) -> P;
+}
```
mitinarseny: This could have been a sigle trait:
```rust
pub trait ProtoTryFrom<P>: Sied {
    fn proto_try_from(value: P) -> anyhow::Result<Self>;
}
```
Into-semantics should be a blanket impl, not vice versa

crates/outlayer/worker/src/main.rs
```diff
@@ -19,6 +26,26 @@ struct AppConfig {
...
+    http_server: HttpServerConfig,
+    grpc: GrpcConfig,
+}
+
+#[serde_with::serde_as]
+#[derive(Deserialize)]
+#[serde(default)]
+struct HttpServerConfig {
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    addr: SocketAddr,
```
mitinarseny, line 38: How about having it optional, and let the OS pick a port otherwise?

crates/outlayer/worker/src/main.rs
```diff
@@ -42,20 +69,44 @@ async fn main() -> Result<()> {
...
+
+    let reflection_service = tonic_reflection::server::Builder::configure()
+        .register_encoded_file_descriptor_set(FILE_DESCRIPTOR_SET)
+        .register_encoded_file_descriptor_set(tonic_health::pb::FILE_DESCRIPTOR_SET)
+        .build_v1()?;
+
+    tracing::info!(addr = %config.http_server.addr, "listening");
+
+    Server::builder()
+        .concurrency_limit_per_connection(config.http_server.concurrency_limit_per_connection)
```
mitinarseny: 1. http?
2. put listening log only after we started serving
3. do you limit concurrent executions per connection for healthchecks and reflections as well?

crates/outlayer/service/src/tower_impl.rs
pityjllk: > Also, maybe .instrument() it with the current span?

i think its automatically inherited whenever its polled right? altohough i fixed instrumentation in OutlayerSerivce::execute ✅ Also since tower here is just a thin wrapper decided that its not worth giving it separate span, wdyt?

crates/outlayer/worker/src/main.rs
pityjllk: not sure tbh, you need to pick some port to bind to right? so it can be shared with the client? also in our case it needs to be exposed in docker compose

crates/outlayer/worker/src/main.rs
pityjllk: >http

yeah in principle there are parameters of http server, the grpc is implemented on top of it

>put listening log only after we started serving

👍 

> do you limit concurrent executions per connection for healthchecks and reflections as well?

noo, these are super lightweight so i dont think they should be limited tbh.

crates/outlayer/service/src/tower_impl.rs
pityjllk: it felt natural to me, like its a service and you can enabled well known "tower" compatibilty for it, although since the request type does not really exists outside of tower context maybe you are right, so decided to move it to grpc

crates/outlayer/grpc/src/convert.rs
pityjllk: these were intentionally kept as two separate traits rather than mirroring the std from/into relationship

the directions arent forward/reverse of the same conversion - theyre two distinct conversions with different fallibility 
- tryFromProto (proto to native) is fallible
- ntoProto (native to proto) is infallible our native types are 

i didnt want to mimic from/into blanket impls at all, its just solution to orphan rule. also these live in the private mod convert so theyre not exposed its just an internal convenience not something we need to shape around std conventions, wdhy?

mitinarseny: Yeah but I was more referring to `Into*<T>` semantics itself: such the trait definition means "there can be many `Into*<P>` impls for a single type". Instead, I suggest to only implement `From*<T>` for these types, while `Into*<P>` can have a blanket impl.

In this case, you can avoid having two traits for it: just a single one would work. If you need to indicate infallibility, then you can do the following:
```rust
pub trait ProtoTryFrom<P>: Sized {
    type Error; // can be `::anyhow::Error` or `::core::convert::Infallible`
    fn proto_try_from(value: P) -> Result<Self, Self::Error>;
}
```

crates/outlayer/worker/src/main.rs
mitinarseny: > in principle there are parameters of http server

In principle, they are parameters of TCP server. Let's name under the final protocol that runs there. We can have a HTTP REST API if needed in the future, so it can be launched on another port, and this config part will be named `http`.

crates/outlayer/worker/src/main.rs
mitinarseny: I was more referring to port `0` meaning "let the OS pick the port", which can be useful for ad-hoc testing of multiple non-conflicting servers running in parallel. But sure, we can add it later

crates/outlayer/worker/src/main.rs
```diff
@@ -42,20 +69,45 @@ async fn main() -> Result<()> {
...
+        .set_service_status("", ServingStatus::Serving)
+        .await;
+
+    let reflection_service = tonic_reflection::server::Builder::configure()
+        .register_encoded_file_descriptor_set(FILE_DESCRIPTOR_SET)
+        .register_encoded_file_descriptor_set(tonic_health::pb::FILE_DESCRIPTOR_SET)
+        .build_v1()?;
+
+    let incoming = TcpIncoming::bind(config.grpc_server.addr).context("bind")?;
+    tracing::info!(addr = %incoming.local_addr().context("local addr")?, "listening");
```
mitinarseny:
```suggestion
    tracing::info!(addr = %incoming.local_addr().context("local addr")?, "gRPC listening");
```

mitinarseny (dismissed):

mitinarseny (approved):

PR #278, opened 2026-05-27
feat: CKD

mitinarseny:
## Summary by CodeRabbit

* **New Features**
  * BLS12-381 child key derivation with public-key verification and decrypt/verify APIs
  * Deterministic MPC helpers for tweak and AppId derivation
  * Multi-version RNG compatibility layer supporting multiple rand_core releases

* **Refactor**
  * MPC KDF reorganized into focused submodules with simplified exports

* **Tests**
  * Added regression/unit tests to lock derivation and verification behavior

* **Chores**
  * Workspace dependency additions and pins updated; runtime libs bumped

[Review Change Stack](https://app.coderabbit.ai/change-stack/near/intents/pull/278?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


crates/rand-compat/Cargo.toml
gilcu3, line 1: Oh this is a nice idea :)

crates/ckd/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+        app_id: &[u8; 32],
+        resp: CkdResponse,
+    ) -> Option<Secret> {
+        let secret = self.decrypt(resp).to_affine();
+
+        if !Self::verify(mpc_public_key, app_id, secret) {
+            return None;
+        }
+
+        Some(secret.to_compressed())
```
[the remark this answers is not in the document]
gilcu3: As long as `verify` makes all the required checks, I don't see why `decrypt` needs to make any check

gilcu3 (commented):
The implementation LGTM. Of course it would be better to have the CKD test vectors you requested, so please follow this https://github.com/near/mpc/issues/3351 so that you are aware once it is implemented

crates/ckd/src/lib.rs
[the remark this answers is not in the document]
mitinarseny: `verify()` is a private method and its usage is enforced whenever you call a public `decrypt_verify()` method

crates/rand-compat/src/lib.rs
mateuszaaa, line 1: i personally think its overkill, almost 400 lines of code that is hard to read cause everthing looks alike and is solving a problem that can be solved in ~20 lines .... 🤔  are you sure?. 

why not  simply accepting rand_core 10 trait at the interface level and implementing single bridge between 0.10 => 0.6. and expect callers to use latest rand core ? and if they are bounded to older version they can implement their own wrapper (so another extra 20 lines and thats it...

crates/ckd/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    /// See <https://github.com/near/mpc/blob/f7a959d2bfd723e92c3bd71a5b60e03d972a2ddb/crates/ckd-example-cli/src/ckd.rs#L128-L129>
+    fn decrypt(&self, resp: CkdResponse) -> G1Projective {
+        resp.big_c - resp.big_y * self.0
+    }
+
+    /// Check that `e(sig, g2) = e(hash_point, mpc_public_key)`
+    ///
+    /// See <https://github.com/near/mpc/blob/f7a959d2bfd723e92c3bd71a5b60e03d972a2ddb/crates/ckd-example-cli/src/ckd.rs#L100-L115>
+    fn verify(mpc_public_key: G2Affine, app_id: &[u8; 32], signature: G1Affine) -> bool {
```
mateuszaaa, line 114: doesnt it makes senseto make it public as well? seems like might fit public api?

crates/kdf/mpc/src/tweak/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+use crate::derive_from_path;
+
+pub type TweakSchema<C> = Derive<ToScalar<C>, Digest<Sha3_256>>;
+
+/// Prepare [`Schema`](defuse_kdf::Schema) for MPC tweak derivation.
+pub fn tweak<C>(predecessor_id: impl AsRef<AccountIdRef>) -> TweakSchema<C>
+where
+    C: NearMpcCurve,
+{
+    ToScalar::<C>::new().derive(derive_tweak(predecessor_id))
```
mateuszaaa, line 22:
```suggestion
    ToScalar::<C>::default().derive(derive_tweak(predecessor_id))
```

crates/kdf/mpc/src/tweak/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+
+#[autoimpl(Debug, Clone, Copy, Default)]
+pub struct ToScalar<C>(PhantomData<C>);
+
+impl<C> ToScalar<C> {
+    #[inline]
+    pub const fn new() -> Self {
+        Self(PhantomData)
+    }
+}
```
mateuszaaa, line 45:
```suggestion

```

mateuszaaa (changes requested):

crates/kdf/mpc/src/tweak/mod.rs
mitinarseny: Benefit of it is `const`

crates/ckd/src/lib.rs
mitinarseny: `signature` here is `secret`, it's only possible to verify if you know the secret. So, I thought, it doesn't make sense to verify and not decrypt

crates/rand-compat/src/lib.rs
mitinarseny: While implementing it, I've seen at least 2-3 crates that do not use latest `rand_core` and which we might use them in the observable future, so it can be re-used there.

mateuszaaa (dismissed):
lgtm

crates/ckd/src/lib.rs
```diff
@@ -0,0 +1,345 @@
...
+        // TODO: cfg(near)
+        Bls12::multi_miller_loop(&[
+            (&resp.big_c, &minus_g2.into()),
+            (&resp.big_y, &self.pk2.into()),
+            (&hp.into(), &mpc_public_key.into()),
+        ])
+        .final_exponentiation()
+        .is_identity()
+        .into()
+    }
```
SimonRastikian, line 193: What is the connection between the link you mention in the comment and this implementation?

mitinarseny: Ooops, forgot to validate the points in the response. Thanks for noticing! Addressed in #279 

As for multi-miller-loop, IIUC that's what `env::bls12381_pairing_check()` [does](https://github.com/near/nearcore/blob/d02248a337d50cda8095a28868ce5add2719c08a/runtime/near-vm-runner/src/logic/bls12381.rs#L332-L338), no?

PR #293, opened 2026-06-25
refactor: `defuse-time` crate

mitinarseny:
## Summary by CodeRabbit

* **New Features**
  * Introduced a unified `Timestamp` type with `borsh`/`serde` adapters and schema support for consistent encoding/decoding.
* **Bug Fixes**
  * Updated all deadline/expiry checks to use consistent “expired/has passed” timestamp comparisons across wallet, escrow, nonce, and intent flows.
* **Refactor**
  * Migrated deadline-related payloads, inspectors, and reporting models to `Timestamp`.
  * Streamlined Borsh utility adapter imports to the direct top-level style.
* **Breaking Changes**
  * Removed the legacy `Deadline` primitive and older Borsh adapter paths/utilities (including chrono/base64 helpers).


crates/primitives/time/src/serde.rs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            fn deserialize_as<D>(deserializer: D) -> Result<Timestamp, D::Error>
+            where
+                D: Deserializer<'de>,
+            {
+                let timestamp = <F as DeserializeAs<'de, $int>>::deserialize_as(deserializer)?;
+                Timestamp::$from(timestamp).ok_or(Overflow).map_err( de::Error::custom)
+            }
+        }
+
```
hlgltvnnk, line 42: Is it possible to make this macro universal? So that you can pass any type, not just Timestamp - and then move it to defuse serde utils

crates/borsh-utils/Cargo.toml
```diff
@@ -1,24 +1,21 @@
...
-chrono = { workspace = true, optional = true }
-
 [features]
-abi = ["borsh/unstable__schema"]
-base64 = ["dep:base64"]
-chrono = ["dep:chrono"]
+schema = ["borsh/unstable__schema"]
+
+[dev-dependencies]
+borsh = { workspace = true, features = ["derive"] }
```
hlgltvnnk, line 21: that is not dev dependency

crates/borsh-utils/src/lib.rs
```diff
@@ -1,4 +1,581 @@
...
+        R: io::Read,
+    {
+        let mut buf = Vec::new();
+        As1::deserialize_as(&mut reader.tee(&mut buf))
+            .or_else(|_| As2::deserialize_as(&mut buf.chain(reader)))
+    }
+}
+
+#[cfg(test)]
+mod tests;
```
hlgltvnnk, line 581: may be move roundtrip_as into defuse-test-utils?

hlgltvnnk (dismissed):

crates/borsh-utils/src/lib.rs
mitinarseny: From the first glance it might sound as a good idea, but when you actually work with `defuse-test-utils`, it becomes a pain, since we didn't put enough efforts to make its API good enough, and then it's harder and harder to make it work. I prefer not to depend on it in newer crates, and, I hope, we can deprecate it eventually

crates/borsh-utils/Cargo.toml
mitinarseny: `borsh/derive` feature is only needed in tests and doctests

crates/primitives/time/src/serde.rs
mitinarseny: Maybe, but I don't think it's worth the effort to generalize it at this point

crates/primitives/time/src/borsh.rs
```diff
@@ -0,0 +1,230 @@
...
+        as_secs,
+        from_secs,
+    }
+
+    pub struct TimestampMilliSeconds: i64 {
+        as_millis,
+        from_millis,
+    }
+
+    pub struct TimestampMicroSeconds: i128 {
```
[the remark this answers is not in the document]
mitinarseny: Yeah, but I removed "default", so users need to specify it manually every time

contracts/defuse/core/src/engine/mod.rs
```diff
@@ -65,7 +65,7 @@ where
         self.inspector.on_deadline(deadline);
 
         // make sure message is still valid
-        if deadline.has_expired() {
+        if deadline < Timestamp::now() {
```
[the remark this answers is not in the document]
mitinarseny: True, but in smart-contract environment `near_sdk::env::block_timestamp()` is always constant within single FunctionCall execution

PR #327, opened 2026-07-27
feat(outlayer-app): granular features

mateuszaaa:
## Summary by CodeRabbit

- **New Features**
  - Added the `near-oa` command-line tool for generating deployment storage data.
  - Supports code hashes provided as base58, hexadecimal, files, or standard input.
  - Outputs single-line JSON with base64-encoded storage keys and values.
  - Added flexible Outlayer App state configuration for administrator, code hash, and code URL.

- **Documentation**
  - Updated installation, command usage, hash format, output, and deployment examples.
  - Clarified the `near-oa` extension workflow and event formatting.


crates/outlayer-app/core/src/lib.rs
```diff
@@ -1,42 +1,42 @@
...
+/// State of an Outlayer App contract
+#[cfg_attr(feature = "parse", derive(Debug))]
+#[derive(Clone, PartialEq, Eq)]
+#[allow(clippy::struct_field_names)]
 pub struct State<'a> {
     pub admin_id: Cow<'a, AccountIdRef>,
     #[cfg_attr(feature = "serde", serde_as(as = "::serde_with::hex::Hex"))]
     pub code_hash: [u8; 32],
     pub code_url: Cow<'a, str>,
+    pub admin_public_key: AdminPublicKey,
```
hlgltvnnk: Does it make sense to have different admins for different purposes? For example, one for updating the code and another for modifying the config - and accordingly to assign different roles to them, so near acl can be used here? But than it would not be part of state init

crates/outlayer-app/core/src/state.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    use super::*;
+
+    #[test]
+    fn immutable_state_uses_zero_implicit_admin() {
+        let admin_public_key = AdminPublicKey::Ed25519(Ed25519PublicKey([0; 32]));
+        let state = State::new_immutable(admin_public_key);
+
+        assert_eq!(state.admin_id.as_ref(), IMMUTABLE_ADMIN_ID);
+        assert_eq!(state.admin_public_key, admin_public_key);
```
hlgltvnnk: failing in case of --no-default-features

hlgltvnnk (changes requested):

crates/outlayer-app/core/src/lib.rs
mateuszaaa: i dont think so the idea is that the owner of the app have control of it. in our case (defuse) its gonna be dao either way

PR #335, opened 2026-08-05
feat: NEP-641 reference impl

mitinarseny:
## Summary by CodeRabbit

* **New Features**
  * Added NEP-641 off-chain authorization with nested paths, access-key and contract resolution, and RPC support.
  * Added off-chain message signing and verification, including Ed25519, WebAuthn, and secp256k1 support.
  * Added wallet SDK support for signing off-chain payloads and wrapping them as extensions.
  * Added wallet operations for managing signature authentication and extensions.
* **Bug Fixes**
  * Strengthened validation for timestamps, paths, signers, chains, permissions, and payloads.
* **API Updates**
  * Split request and off-chain signature APIs and updated authorization resolution methods.


contracts/wallet/src/error.rs
```diff
@@ -14,17 +14,27 @@ pub enum ContractError {
...
     InvalidSignature,
 
     #[error("invalid signer_id: {0}")]
     InvalidSignerId(AccountId),
 
-    #[error("insufficient attached deposit")]
-    InsufficientDeposit,
+    #[cfg(feature = "json")]
+    #[error("JSON: {0}")]
+    JSON(#[from] serde_json::Error),
```
frol, line 37: This naming is misleading - it is not a JSON error, it is an error of serializing / deserializing JSON data. So it would be better if named something like "SerializationError" or "InvalidDataFormat"

crates/signatures/nep641/src/resolver/error.rs
```diff
@@ -0,0 +1,58 @@
...
+    AccessKey(#[from] AccessKeyError),
+
+    #[error(transparent)]
+    Contract(#[from] ContractError),
+
+    #[error("resolved payload is invalid: expected: {}, got: {}", .expected, .payload)]
+    InvalidPayload { payload: String, expected: String },
+
+    #[error("JSON: {0}")]
+    JSON(#[from] serde_json::Error),
```
frol, line 33: Same comment as above regarding the "JSON" name

contracts/wallet/src/contract.rs
```diff
@@ -341,6 +329,85 @@ where
...
+
+                // check signer_id
+                if msg.signer_id != env::current_account_id() {
+                    return Err(Error::InvalidSignerId(msg.signer_id));
+                }
+
+                // check reversed path
+                if !msg.path.iter().eq(path.iter().rev()) {
+                    return Err(Error::InvalidPath);
+                }
```
frol: We discussed it offline and the reasoning for having these two lists in different order is currently the following:

1. `msg.path` goes from the current signer to the top-level signer per the Ledger guidelines
2. `path` arg goes in the opposite direction due to resolver efficiency (Vec::append is cheap)

That said, I don't like when seemingly similar things behave differently. So I suggest we reconsider using Vec on the resolver or teach it to serialize in the reverse order, so that on the contract side / on the NEP side the order is always the same.

contracts/wallet/src/error.rs
mitinarseny: JSON is a serialization format itself and `serde_json::Error` is used for both serialization and deserialization, which propagates through `?` on call sites and integrates nicely with `thiserror`'s `#[from]`. Lets leave it for future optimization / refactoring if ever needed

PR #341, opened 2026-08-08
Add unit tests for defuse-map-utils' DefaultMap cleanup semantics

ayushsingh82:
## Summary

`crates/map-utils` has zero test files across ~1,150 lines. `cleanup.rs` in particular implements `DefaultMap::entry_or_default`, which relies on `Drop` to decide whether an entry actually stays in the underlying map:

- A **vacant** entry only gets inserted if it's mutated away from `Default` before being dropped.
- An **occupied** entry gets removed if it's mutated back to `Default` before being dropped.
- An explicit `.remove()` bypasses that `Drop` logic entirely (for a vacant entry it just discards the never-inserted default; for an occupied one it removes and returns the current value).

That logic was covered only by two doctests exercising a single happy path. This adds a proper unit test suite (`crates/map-utils/src/cleanup/tests.rs`) covering:

- Vacant entry dropped with/without mutation (insert vs. no-op)
- Vacant entry mutated then reverted back to default (no insert)
- Occupied entry dropped with/without mutation, and mutated to/away from default (remove vs. keep vs. update)
- Explicit `.remove()` on both vacant and occupied entries
- `.key()` on both entry kinds
- `Deref`/`DerefMut` read/write through
- Independence between multiple keys in the same map

No behavior changes — pure test addition.

## Verification
- `cargo test -p defuse-map-utils --lib`: 13 passed
- `cargo test -p defuse-map-utils --doc`: 3 passed (existing doctests, unaffected)
- `cargo clippy -p defuse-map-utils --all-targets -- -D warnings`: clean
- `cargo fmt -p defuse-map-utils -- --check`: clean
- Note: `cargo clippy -p defuse-map-utils --all-targets --all-features` (i.e. including the `near` feature) fails to compile locally on unrelated grounds — `near-sdk` v5.29.0 refuses to build outside a small set of cfgs (`cargo near build`, `cfg(clippy)`, etc.) — pre-existing and unrelated to this change; the `near` feature isn't touched here.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for default map entry behavior across supported map types.
  * Tested insertion, mutation, removal, key access, dereferencing, and independent keys.
  * Verified that existing entries are preserved and updated correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1617,6 +1617,7 @@ version = "0.1.0"
 dependencies = [
  "impl-tools 0.12.0",
  "near-sdk",
+ "rstest",
 ]
 
 [[package]]
--- a/crates/map-utils/Cargo.toml
+++ b/crates/map-utils/Cargo.toml
@@ -16,3 +16,7 @@ near-sdk = { workspace = true, optional = true }
 
 [features]
 near = ["dep:near-sdk"]
+
+[dev-dependencies]
+near-sdk = { workspace = true, features = ["unit-testing"] }
+rstest.workspace = true
--- a/crates/map-utils/src/cleanup.rs
+++ b/crates/map-utils/src/cleanup.rs
@@ -273,3 +273,6 @@ where
         }
     }
 }
+
+#[cfg(test)]
+mod tests;
--- a/crates/map-utils/src/cleanup/tests.rs
+++ b/crates/map-utils/src/cleanup/tests.rs
@@ -1,0 +1,117 @@
+use std::collections::HashMap;
+
+use super::*;
+
+#[test]
+fn vacant_entry_dropped_without_mutation_does_not_insert() {
+    let mut m: HashMap<&str, i32> = HashMap::new();
+    let _ = m.entry_or_default("a");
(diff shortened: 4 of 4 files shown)


mitinarseny:
Hey, thanks for the PR!
Would you mind using `rstest` crate to make these tests generic and reused for all: `HashMap`, `BTreeMap`, and near collections (if `near` feature is enabled)?

ayushsingh82:
Done — the suite now runs each test case against `HashMap`, `BTreeMap`, and (behind the `near` feature) `near_sdk::store::IterableMap` via `rstest`'s `#[case]`, instead of only `HashMap`.

Verification:
- `cargo test -p defuse-map-utils --lib`: 26 passed (13 tests × 2 map kinds)
- `cargo test -p defuse-map-utils --lib --features near`: 39 passed (13 tests × 3 map kinds)
- `cargo test -p defuse-map-utils --doc`: 3 passed
- `cargo clippy -p defuse-map-utils --all-targets --all-features -- -D warnings`: clean
- `cargo fmt -p defuse-map-utils -- --check`: clean

@mitinarseny

PR #344, opened 2026-08-08
Add unit tests for defuse-num-utils' checked arithmetic traits

ayushsingh82:
## Summary

`crates/num-utils` (~190 lines) implements `CheckedAdd`, `CheckedSub`, `CheckedMul`, `CheckedDiv`, and `CheckedMulDiv` across every integer width (`u8..u128`, `i8..i128`), with zero test coverage. It's a dependency of `defuse-fees` and `defuse-decimal`'s ordering/arithmetic, so a rounding or overflow-narrowing bug here silently affects fee and price math.

This adds unit test coverage (`rstest`-based, following the pattern used elsewhere in the workspace, e.g. `crates/primitives/decimal/src/ops.rs`) for:

- **`CheckedAdd`/`CheckedSub`**: same-type, `unsigned + signed`, and `signed + unsigned` combinations, plus overflow/underflow boundaries at both the narrow (`i8`/`u8`) and wide (`i128`/`u128`) ends.
- **`CheckedDiv`**: division by zero, exact division, and rounding-up behavior of `checked_div_ceil`.
- **`CheckedMul`**: overflow, and that the default `checked_mul_ceil` forwards to `checked_mul` (no type currently overrides it).
- **`CheckedMulDiv`**: division by zero, the case where the widened intermediate product doesn't overflow but the *final narrowed* result no longer fits back into the base type, and — for the `i128`/`BInt<4>` path — euclidean vs. truncating rounding on negative operands (`checked_mul_div` truncates toward zero, `checked_mul_div_euclid` rounds toward negative infinity).

All expected values were cross-checked against the actual `bnum`-backed implementation in a standalone harness before being hardcoded into the tests, rather than hand-computed.

No behavior changes — pure test addition.

## Verification
- `cargo test -p defuse-num-utils --lib`: 75 passed
- `cargo test -p defuse-num-utils --doc`: 0 (no doctests in this crate)
- `cargo clippy -p defuse-num-utils --all-targets -- -D warnings`: clean
- `cargo fmt -p defuse-num-utils -- --check`: clean
- `cargo check -p defuse-fees -p defuse-decimal`: unaffected (test-only change)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1790,6 +1790,7 @@ name = "defuse-num-utils"
 version = "0.1.0"
 dependencies = [
  "bnum",
+ "rstest",
 ]
 
 [[package]]
--- a/crates/num-utils/Cargo.toml
+++ b/crates/num-utils/Cargo.toml
@@ -9,3 +9,6 @@ repository.workspace = true
 
 [dependencies]
 bnum.workspace = true
+
+[dev-dependencies]
+rstest.workspace = true
--- a/crates/num-utils/src/add_sub.rs
+++ b/crates/num-utils/src/add_sub.rs
@@ -74,3 +74,89 @@ impl_checked!(u16, i16);
 impl_checked!(u32, i32);
 impl_checked!(u64, i64);
 impl_checked!(u128, i128);
+
+#[cfg(test)]
+mod tests {
+    use core::fmt::Debug;
+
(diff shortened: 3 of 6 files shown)


ayushsingh82:
@mitinarseny would appreciate your review on this one too, thanks!

PR #350, opened 2026-08-26
test(nep641): add canonical test vectors for OffchainMessage hashing

jatinsahijwani:
Adds a published set of canonical test vectors for `OffchainMessage` hashing, plus a test asserting the implementation reproduces them.

Why: I am implementing the TypeScript resolver that the NEP lists as TBD, and there was no published conformance suite to check against. In the forum thread a wallet team specifically asked for shared cross-language fixtures in a commonly referenced location, on the grounds that cross-language consistency matters here and downstream implementations should not each invent their own.

These sit next to the code that defines them rather than in a downstream package, and the accompanying test means they cannot drift from the implementation.

Coverage includes empty and non-empty `path`, subsecond timestamps with and without trailing zeros, and the existing doctest cases.

Forum discussion: https://gov.near.org/t/building-the-typescript-resolver-for-nep-641-currently-listed-as-tbd/42479

Happy to change the format or location if the crate would prefer something different.

## Summary by CodeRabbit

* **Documentation**
  * Added documentation for NEP-641 off-chain authorization test vectors and message digest formatting.

* **Tests**
  * Added validation for published off-chain message hash vectors, including domain separators and message variants.
  * Added coverage for empty, single-hop, nested resolver paths, and equivalent timestamps with different trailing-zero formats.

### DIFF
--- a/crates/signatures/nep641/Cargo.toml
+++ b/crates/signatures/nep641/Cargo.toml
@@ -109,6 +109,7 @@ resolver = [
 [dev-dependencies]
 defuse-nep641 = { path = ".", features = ["abi", "borsh", "digest", "serde", "near-kit", "resolver", "tracing"] }
 
+hex.workspace = true
 hex-literal.workspace = true
 near-kit = { workspace = true, features = ["sandbox"] }
 tokio = { workspace = true, features = ["macros", "rt"] }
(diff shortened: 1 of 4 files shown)


frol (approved):
Thanks for the conformance tests! We will need to add more of these in the future, so do not hesitate to expand the test suite.

PR #355, opened 2026-09-23
impl: lazy state migration

commit4ikus:


contracts/poa/factory/src/contract/versioned/tests.rs
```diff
@@ -0,0 +1,90 @@
...
+    tokens.insert("usdc".to_string());
+
+    ContractStorageV0 {
+        tokens,
+        bridge_token_storage_deposit_required: NearToken::from_millinear(500),
+    }
+}
+
+#[test]
+fn legacy_upgrade() {
```
karim-en: The unit tests are almost useless for this kind of migration, let's use integration tests and download the previous wasm from github and apply the upgrade.
Check this example 
https://github.com/Near-One/btc-bridge/blob/omni-main/contracts/satoshi-bridge/tests/test_upgrade.rs

karim-en (changes requested):

contracts/poa/factory/src/contract/versioned/tests.rs
commit4ikus: Agreed

Added migration/upgrade tests (tests/src/tests/poa/upgrade.rs).

Run:
cargo test -p defuse-tests --no-default-features --features poa test_poa_factory_upgrade_from_legacy -- --nocapture

Log example: 
image

works

commit4ikus: upd: added test for ft_wthdraw (should be checked after update too)

image

